PR #78, opened 2017-11-21
STYLE: Remove superfluous calls to "find_package(Slicer ...)"

jcfr:


DoseVolumeHistogram/SubjectHierarchyPlugins/qSlicerSubjectHierarchyDoseVolumeHistogramPlugin.cxx
```diff
@@ -349,13 +349,13 @@ vtkIdType qSlicerSubjectHierarchyDoseVolumeHistogramPlugin::getChartShItemForDvh
   if (!scene)
   {
     qCritical() << Q_FUNC_INFO << ": Invalid MRML scene";
-    return NULL;
+    return 0;
```
cpinter, line 352: Why are these changes necessary? NULL is everywhere in SlicerRT (and in Slicer) and it works well. Actually in SlicerRT I consider this part of the coding convention.

jcfr: It was reporting a warning. 0 if for number, NULL for pointer.

lassoan: For me, 0 is as bad as NULL (but with NULL the coder's intent is a bit more clear). Fortunately, nullptr is introduced in cxx11 - we should just switch to that.

jcfr: To clarify, the return type is `vtkIdType`, I don't think `nullptr` would work either.

lassoan: Sure, if it's vtkIdType then it should be 0. Or, even nicer: vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID

cpinter: Ah OK. If it's a number, then of course it should be 0. vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID was the original constant for invalid SH ID, but then we changed it to be 0 just because it gives more concise, easier to read (and write) code. 0 is fine with me!

If it was a pointer, then my feelings coincide with Andras'. I use NULL mostly for clarity (it's then obvious it's a pointer). I also agree with the eventual shift to nullptr.

In summary no need to change anything here :) Thanks for the answers

PR #81, opened 2018-01-29
BUG: If slice thickness cannot be calculated, default to image slice spacing

Sunderlandkyl:


DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
```diff
@@ -938,6 +941,10 @@ bool vtkSlicerDicomRtImportExportModuleLogic::vtkInternal::LoadRtStructureSet(vt
         scene->AddNode(segmentationDisplayNode);
         segmentationNode->SetAndObserveDisplayNodeID(segmentationDisplayNode->GetID());
         segmentationDisplayNode->SetBackfaceCulling(0);
+
+        // Setup default slice thickness for planar contour to closed surface conversion
+        double defaultSliceThickness = referencedVolumeSpacing[2];
```
cpinter: Please initialize to 0,0

cpinter:
@Sunderlandkyl Can you please fix this? Thanks!
http://slicer.cdash.org/viewBuildError.php?buildid=1189651

Sunderlandkyl:
@cpinter Fixed in 1aac86b9c0b29c6e7501477f39da2dec17540871

I noticed something odd however. Getting the spacing of the associated volume still fails sometimes since the image is not always found.

Specifically, this line: https://github.com/SlicerRt/SlicerRT/blob/master/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx#L915-L916
Will return vtkMRMLSubjectHierarchyNode::INVALID_ITEM_ID occasionally.

Is the load order of dicom static, or is the volume sometimes loaded after the RTStruct?

lassoan:
Loading order is random (or the user may choose not to load it at all). However, you may retrieve information about the volume from the DICOM database.

cpinter:
There's also a way to deal with the random loading order for the RT image. You may do the same for this case.

PR #82, opened 2018-01-31
BUG: Calculate default slice thickness if volume has not been loaded

Sunderlandkyl:
Uses GDCM::IPPSorter to calculate slice spacing.
- From gdcmIPPSorter.h:
  > ALL slices are taken into account, if one slice is missing then ZSpacing will be set to 0 since the spacing  will not be found to be regular along the Series"

There are 2 adjustable parameters that are left as default:
- ZSpacingTolerance :: 1e-6
- DirectionCosinesTolerance :: 0

### DIFF
--- a/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
+++ b/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
@@ -112,6 +112,9 @@
 // ITK includes
 #include <itkImage.h>
 
+// GDCM includes
+#include <gdcmIPPSorter.h>
+
 // DICOMLib includes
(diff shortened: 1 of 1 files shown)


cpinter:
There is a quite complex and comprehensive logic for calculating slice thickness in vtkPlanarContourToRibbonModelConversionRule. Can you please explain why this is needed and why use a new implementation?

Sunderlandkyl:
In order to calculate the default spacing in the instances where there were too few contours to calculate spacing, the default contour spacing is set to be the same as the image volume slice thickness for planar contour to closed surface conversion.

If the image volume hasn't been loaded yet, then there needs to be a fallback method, which in this case is using the spacing calculated from Image Position (Patient) and Image Orientation (Patient) in the Dicom.

cpinter:
Thanks!

PR #83, opened 2018-03-07
ENH: Added util function to call ctcreate program

anna-ilina:
Add utility function to call ctcreate program from slicer. Ctcreate converts a DICOM phantom to a ctcreate phantom, which is the input needed for DOSXYZnrc.

### DIFF
--- a/ExternalBeamPlanning/DoseEngines/Python/OrthovoltageDoseEngineUtil.py
+++ b/ExternalBeamPlanning/DoseEngines/Python/OrthovoltageDoseEngineUtil.py
@@ -1,6 +1,4 @@
-import sys
-from os import listdir
-from os.path import isfile, join
+import os
 import vtk, qt, ctk, slicer
 import logging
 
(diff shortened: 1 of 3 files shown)


ExternalBeamPlanning/DoseEngines/Python/OrthovoltageDoseEngineUtil.py
```diff
@@ -0,0 +1,127 @@
...
+
+  # CT Record 6 information about material (for i=1 to num_material)
+  outFile.write("AIR521ICRU\n")
+  outFile.write("-974, 0.001, 0.044\n")
+  outFile.write("LUNG521ICRU\n")
+  outFile.write("-724, 0.044, 0.302\n")
+  outFile.write("ICRUTISSUE521ICRU\n")
+  outFile.write("101, 0.302, 1.101\n")
+  outFile.write("ICRPBONE521ICRU\n")
+  outFile.write("1976, 1.101, 2.088")
```
cpinter, line 94: This is a lot of magic numbers. Please provide much more explanation in the comment about what these are and why they are not configurable.

cpinter:
I thought the change will be bigger. Considering it's a quite small one, can you please squash the commits after all? Thanks!

Also please see comment in the diff.

cpinter:
Thanks!

PR #89, opened 2018-06-19
Missing include for QButtonGroup

auneri:
Necessary (at least) with Qt 5.11 on macOS 10.12.


lassoan (approved):
Looks good, thank you!

PR #90, opened 2018-07-30
BUG: Fix PNGs with mogrify so Slicer doesn't give a libpng warning

tavaughan:
Command used:
find -name "*.png" -exec 'mogrify' {} \;

Some background:
https://stackoverflow.com/questions/22745076/libpng-warning-iccp-known-incorrect-srgb-profile


cpinter:
Excellent, thanks for fixing this annoying warning!

PR #93, opened 2018-10-23
ENH: Add dose surface histogram functionality

Sunderlandkyl:
Adds the option to calculate dose surface histogram instead of dose volume histogram.
- Checkbox for toggling DSH/DVH (DVH by default)
- Option on node for toggling inside/outside surface (Inside by default)

image


Sunderlandkyl:
@cpinter I'm going to add some tests as well. Probably one for the ENT phantom and one for the prostate phantom, unless you have any other suggestions.

Sunderlandkyl:
If there is an exported DSH from some other platform (Eclipse/CERR, etc.), I think it would be good to compare against that as well.

cpinter:
I'll look into baselines.

Sunderlandkyl:
Tests added.
- 2 Prostate phantom (inside/outside)
- 2 ENT phantom (inside/outside)

My experience so far has suggested that the volume of the structure (meaning the volume of the surface voxels), can vary by quite a bit between DSH calculations, even for the same structure. 

Likely this has to do with voxel size (larger voxels -> thicker shells), but I thought I'd bring it up.

cpinter:
Thanks a lot, Kyle! I'll need to see these varying sizes so that I can form an opinion. Probably tomorrow.

PR #104, opened 2019-04-18
Fix compilation warnings

jcfr:


cpinter:
Great, thank you!

jcfr:
As a side note, these vtkMRMLRoomsEyeViewNode attributes are not used:

```
  int ApplicatorHolderVisibility;
  int ElectronApplicatorVisibility;
```

cpinter:
That module has bene in progress for a while, so I'm not surprised. I took it out from the experimental modules so that the factory runs its tests - one was failing for a while - and because it's a neat module nevertheless.
Maybe I'll just comment out those variables.

cpinter:
@jcfr FYI I just added the initialization, XML, Copy etc for these variables
https://github.com/SlicerRt/SlicerRT/commit/e14c55ed03ef57183689199686d3f7c7d5dda4a1

PR #116, opened 2019-09-30
BUG: Fix failure in BatchStructureSetConversion

Sunderlandkyl:
ConvertStructureSetToLabelmap requires use_ref_image argument, which was not specified in the test.
This commit sets use_ref_image=True for the test.


cpinter:
Thanks, @Sunderlandkyl ! LGTM

PR #117, opened 2019-10-01
ENH: Add support for merged binary labelmap segments

Sunderlandkyl:
Binary labelmap representations can now support multiple segments in the same vtkOrientedImageData.
This commit replaces some instances of segment->GetRepresentation("Binary labelmap") with segmentationNode->GetBinaryLabelmapRepresentation(), or the use of vtkImageThreshold to isolate the binary labelmaps for individual segments.

To be merged once https://github.com/Slicer/Slicer/pull/1221 has been merged.

### DIFF
--- a/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
+++ b/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
@@ -2021,8 +2021,8 @@ std::string vtkSlicerDicomRtImportExportModuleLogic::ExportDicomRTStudy(vtkColle
         vtkSegment* segment = segmentationNode->GetSegmentation()->GetSegment(*segmentIdIt);
 
         // Get binary labelmap representation
-        vtkOrientedImageData* binaryLabelmap = vtkOrientedImageData::SafeDownCast(
-          segment->GetRepresentation(vtkSegmentationConverter::GetSegmentationBinaryLabelmapRepresentationName()) );
+        vtkNew<vtkOrientedImageData> binaryLabelmap;
+        segmentationNode->GetBinaryLabelmapRepresentation(segmentID, binaryLabelmap);
         if (!binaryLabelmap)
         {
           error = "Failed to get binary labelmap representation from segment " + segmentID;
--- a/DoseComparison/Logic/vtkSlicerDoseComparisonModuleLogic.cxx
+++ b/DoseComparison/Logic/vtkSlicerDoseComparisonModuleLogic.cxx
@@ -231,8 +231,8 @@ std::string vtkSlicerDoseComparisonModuleLogic::ComputeGammaDoseDifference(vtkMR
       return errorMessage;
     }
     // Get segment binary labelmap
-    vtkOrientedImageData* maskSegmentLabelmap = vtkOrientedImageData::SafeDownCast( segmentationCopy->GetSegment(maskSegmentID)->GetRepresentation(
-      vtkSegmentationConverter::GetSegmentationBinaryLabelmapRepresentationName() ) );
+    vtkNew<vtkOrientedImageData>  maskSegmentLabelmap;
+    maskSegmentationNode->GetBinaryLabelmapRepresentation(maskSegmentID, maskSegmentLabelmap);
 
     // Apply parent transformation nodes if necessary
     if ( maskSegmentationNode->GetParentTransformNode()
--- a/DoseVolumeHistogram/Logic/vtkSlicerDoseVolumeHistogramModuleLogic.cxx
+++ b/DoseVolumeHistogram/Logic/vtkSlicerDoseVolumeHistogramModuleLogic.cxx
@@ -235,6 +235,9 @@ std::string vtkSlicerDoseVolumeHistogramModuleLogic::ComputeDvh(vtkMRMLDoseVolum
   fixedOversamplingValueStream << this->DefaultDoseVolumeOversamplingFactor;
   segmentationCopy->SetConversionParameter( vtkClosedSurfaceToBinaryLabelmapConversionRule::GetOversamplingFactorParameterName(),
     parameterNode->GetAutomaticOversampling() ? "A" : fixedOversamplingValueStream.str().c_str() );
+  // We don't want to try to merge the labelmaps since if they have different oversampling factors, they would conflict.
+  // Could perhaps leave the labelmaps merged if there is a performance increase, but for now merging will be disabled for DVH calculation.
+  segmentationCopy->SetConversionParameter(vtkClosedSurfaceToBinaryLabelmapConversionRule::GetCollapseLabelmapsParameterName(), "0");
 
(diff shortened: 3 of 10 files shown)


DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
```diff
@@ -2021,8 +2021,8 @@ std::string vtkSlicerDicomRtImportExportModuleLogic::ExportDicomRTStudy(vtkColle
         vtkSegment* segment = segmentationNode->GetSegmentation()->GetSegment(*segmentIdIt);
 
         // Get binary labelmap representation
-        vtkOrientedImageData* binaryLabelmap = vtkOrientedImageData::SafeDownCast(
-          segment->GetRepresentation(vtkSegmentationConverter::GetSegmentationBinaryLabelmapRepresentationName()) );
+        vtkSmartPointer<vtkOrientedImageData> binaryLabelmap = vtkSmartPointer<vtkOrientedImageData>::Take(
+          segmentationNode->GetBinaryLabelmapRepresentation(segmentID));
```
cpinter: Is there any documentation in the GetBinaryLabelmapRepresentation and similar functions that now the returned object needs to be deleted?

DoseComparison/Logic/vtkSlicerDoseComparisonModuleLogic.cxx
```diff
@@ -231,8 +231,9 @@ std::string vtkSlicerDoseComparisonModuleLogic::ComputeGammaDoseDifference(vtkMR
       return errorMessage;
     }
     // Get segment binary labelmap
-    vtkOrientedImageData* maskSegmentLabelmap = vtkOrientedImageData::SafeDownCast( segmentationCopy->GetSegment(maskSegmentID)->GetRepresentation(
-      vtkSegmentationConverter::GetSegmentationBinaryLabelmapRepresentationName() ) );
+
```
cpinter: If this is an extra empty line, please remove.

DoseVolumeHistogram/Logic/vtkSlicerDoseVolumeHistogramModuleLogic.cxx
```diff
@@ -326,8 +329,24 @@ std::string vtkSlicerDoseVolumeHistogramModuleLogic::ComputeDvh(vtkMRMLDoseVolum
     vtkSegment* segment = segmentationCopy->GetSegment(*segmentIdIt);
 
     // Get segment labelmap
-    vtkOrientedImageData* segmentLabelmap = vtkOrientedImageData::SafeDownCast( segment->GetRepresentation(
+    vtkSmartPointer<vtkOrientedImageData> segmentLabelmap = vtkOrientedImageData::SafeDownCast( segment->GetRepresentation(
```
cpinter, line 332: Same here, it's an important (and somewhat dangerous) change that now the returned object needs to be owned. Must be documented well, in GetRepresentation as well.

cpinter:
Minor comments, otherwise looks good (I assume you ran the tests and at least tried the DVH module manually with different settings). Can be committed after the Slicer PR went in. Thank you!

DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
Sunderlandkyl: I added this comment to the header file documentation:
https://github.com/Slicer/Slicer/blob/c9203a9f13fdd727965dd7ec6cbb7623863aaa76/Libs/MRML/Core/vtkMRMLSegmentationNode.h#L210

DoseVolumeHistogram/Logic/vtkSlicerDoseVolumeHistogramModuleLogic.cxx
Sunderlandkyl: This one doesn't need to be owned (vtkSegment::GetRepresentation() still returns a pointer to the original representation in the segment. vtkSegmentationNode::GetBinaryLabelmap() returns a new instance.

Sunderlandkyl:
Thanks for reviewing! I ran all of the SlicerRT tests and everything passed.

cpinter:
If you haven't done it already, please do some manual testing in DVH, trying all options. Thanks!

DoseVolumeHistogram/Logic/vtkSlicerDoseVolumeHistogramModuleLogic.cxx
cpinter: Oh, right. There's no Take

DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
lassoan: The risk of misuse is so high that documentation is not enough. Factory methods must start with a very clear name (usually "Create") but some developers even missed those, so now when we need to assign ownership we always use `GetSomething(vtkSomeObject *objCreatedByCaller)` pattern.

If changing the method signature would cause too much inconvenience for the users then these labelmap representations could be cached in the segmentation node or object (similarly to how returned vectors are cached in classes to allow simple _spacing=img->GetSpacing()_ syntax)

cpinter: I usually use this kind of signature in such cases:

```
vtkNew<vtkOrientedImageData> binaryLabelmap;
segmentationNode->GetBinaryLabelmapRepresentation(segmentID, binaryLabelmap);
```

Sunderlandkyl: Ok, I'll change the function signature in the segmentation node.

Do you want me to change the GetClosedSurfaceRepresentaion signature for consistency?
It is not currently a factory method.

Sunderlandkyl:
The function signature for GetBinaryLabelmapRepresentation and GetClosedSurfaceRepresentation is now:

     virtual void GetBinaryLabelmapRepresentation(const std::string segmentId, vtkOrientedImageData* outputBinaryLabelmap);

cpinter:
Thanks! I think it will be safer this way

lassoan:
@cpinter Do you use this same branch for Slicer-4.10?

lassoan:
Merged in https://github.com/SlicerRt/SlicerRT/commit/2f77010cf96d6d9fef27c8bf02ca319810e9555a. Added #ifdefs to make it work with both Slicer-4.10 and Slicer-4.11 - I did not have a chance to test it, so it would be great if @Sunderlandkyl you could have a look at the dashboard tomorrow and/or test if it works as expected.

cpinter:
Thanks for merging!
@lassoan There has been no need for a separate branch so far. The ifdefs are a good solution, but I wouldn't have minded fixing to the commit before this one; the next stable is around the corner.

PR #120, opened 2019-10-30
initial MLC support

MichaelColonel:
RTPlan only

### DIFF
--- a/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
+++ b/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
@@ -1740,36 +1740,6 @@ void vtkSlicerDicomRtReader::GetBeamLeafJawPositions(unsigned int beamNumber, do
 }
 
 //----------------------------------------------------------------------------
-double vtkSlicerDicomRtReader::GetBeamSourceToJawDistanceX(unsigned int beamNumber)
-{
-  vtkInternal::BeamEntry* beam=this->Internal->FindBeamByNumber(beamNumber);
-  if (beam)
-  {
-    return beam->SourceToJawDistance[0];
-  }
-  else
-  {
-    vtkErrorMacro("GetBeamSourceToJawDistanceX: Unable to find beam of number" << beamNumber);
-    return -1.0;
-  }
-}
-
-//----------------------------------------------------------------------------
-double vtkSlicerDicomRtReader::GetBeamSourceToJawDistanceY(unsigned int beamNumber)
-{
-  vtkInternal::BeamEntry* beam=this->Internal->FindBeamByNumber(beamNumber);
-  if (beam)
-  {
-    return beam->SourceToJawDistance[1];
-  }
-  else
-  {
-    vtkErrorMacro("GetBeamSourceToJawDistanceY: Unable to find beam of number" << beamNumber);
-    return -1.0;
-  }
-}
-
-//----------------------------------------------------------------------------
 bool vtkSlicerDicomRtReader::GetBeamMultiLeafCollimatorPositionsX( unsigned int beamNumber, 
   std::vector<double>& pairBoundaries, std::vector<double>& leafPositions)
 {
--- a/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
+++ b/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
@@ -83,7 +83,7 @@ public:
     }
     /// Source to BeamLimitingDevice (MLC) distance for RTPlan
     /// or Isocenter to BeamLimitingDevice (MLC) distance for RTIonPlan
-    double Distance;
+    double Distance; // loaded, but not used
     unsigned int NumberOfLeafJawPairs;
     /// MLC position boundaries: Raw DICOM values
     std::vector<double> LeafPositionBoundary;
@@ -155,7 +155,9 @@ public:
     double LeafJawPositions[2][2];
 
     /// Jaw distance: X and Y distance from souce with isocenter as origin
-    double SourceToJawDistance[2]; // [0] - X or ASYMX, [1] - Y or ASYMY
+    /// SourceToJawDistance[0] - X or ASYMX
+    /// SourceToJawDistance[1] - Y or ASYMY
+    double SourceToJawDistance[2]; // loaded, but not used
 
     /// MLC parameters: Raw DICOM values
     /// MultiLeafCollimator[0] - MLCX parameters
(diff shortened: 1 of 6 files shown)


cpinter:
Thanks a lot for making the separate PR for the MLCs!

I'll add comments in the code.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -128,6 +148,8 @@ class vtkSlicerDicomRtReader::vtkInternal
     double BeamLimitingDeviceAngle;
     /// Jaw positions: X and Y positions with isocenter as origin (e.g. {{-50,50}{-50,50}} )
     double LeafJawPositions[2][2];
+    /// MLC positions: Raw DICOM values [0] - MLCX, [1] - MLCY
+    std::vector<double> LeafJawPositionsMLC[2];
```
cpinter, line 180: You added the class BeamLimitingDeviceEntry, which is quite nice, but then you add this array of vectors in the beam, which seems a bit disconnected. I'd just like to understand...

So the BeamLimitingDeviceEntry is the general description of one MLC array, and these two vectors are the actual positions of the leaves for the beam, correct? How do we know what corresponds to what? Simply assume that the first leaf position vector corresponds to the first MLC device?

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -558,6 +580,62 @@ void vtkSlicerDicomRtReader::vtkInternal::LoadRTPlan(DcmDataset* dataset)
...
+                    boundY.resize(leafPositionBoundaries.size());
+                    std::copy( leafPositionBoundaries.begin(), leafPositionBoundaries.end(), boundY.begin());
+                    vtkDebugWithObjectMacro( this->External, "LoadRTPlan: MLCY number of leaf/jaw pairs: " << nofPairs);
+                  }
+                }
+              }
+            }
+          }
+        } while (rtBeamLimitingDeviceSequence.gotoNextItem().good());
+      }
```
cpinter, line 674: Very nice code btw, thanks for paying close attention to style!

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -1614,6 +1787,66 @@ void vtkSlicerDicomRtReader::GetBeamLeafJawPositions(unsigned int beamNumber, do
...
+//----------------------------------------------------------------------------
+bool vtkSlicerDicomRtReader::GetBeamMultiLeafCollimatorPositionsX( unsigned int beamNumber, 
+  std::vector< std::pair< double, double> >& leafPairPositions)
+{
+  vtkInternal::BeamEntry* beam = this->Internal->FindBeamByNumber(beamNumber);
+  unsigned int& pairs = this->Internal->MultiLeafCollimator[0].NumberOfLeafJawPairs;
+
+  leafPairPositions.clear();
+  if (beam && ((pairs * 2) == beam->LeafJawPositionsMLC[0].size())) {
+    for ( unsigned int i = 0; i < pairs; ++i)
```
cpinter: Please use curly braces even if the body is one line. If someone adds a second line, they will spend half a day before realizing why it's not included in the loop. Thanks!

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -66,6 +66,26 @@ class vtkSlicerDicomRtReader::vtkInternal
...
+    }
+    /// Source to BeamLimitingDevice (MLC) distance for RTPlan
+    /// or Isocenter to BeamLimitingDevice (MLC) distance for RTIonPlan
+    double Distance;
+    unsigned int NumberOfLeafJawPairs;
+    /// MLC position boundaries: Raw DICOM values
+    std::vector<double> LeafPositionBoundary;
+  };
+  /// [0] - MLCX, [1] - MLCY
+  BeamLimitingDeviceEntry MultiLeafCollimator[2];
```
cpinter: Where is this information filled? I cannot find it in this PR.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -558,6 +580,62 @@ void vtkSlicerDicomRtReader::vtkInternal::LoadRTPlan(DcmDataset* dataset)
...
+                  double& distanceY = MultiLeafCollimator[1].Distance;
+                  unsigned int& pairsX = MultiLeafCollimator[0].NumberOfLeafJawPairs;
+                  unsigned int& pairsY = MultiLeafCollimator[1].NumberOfLeafJawPairs;
+                  std::vector<double>& boundX = MultiLeafCollimator[0].LeafPositionBoundary;
+                  std::vector<double>& boundY = MultiLeafCollimator[1].LeafPositionBoundary;
+                  if (deviceType == "MLCX")
+                  {
+                    distanceX = sourceToMLCDistance;
+                    pairsX = nofPairs;
+                    boundX.resize(leafPositionBoundaries.size());
```
cpinter, line 657: As far as I understand this is where you fill the Internal->MultiLeafCollimator[n] BeamLimitingDeviceEntries. But you have only 2 arrays, and this code here is in a loop. Is it always the same, or how it is not a problem that it gets overwritten at every subsequent run of the loop?

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: Correct. BeamLimitingDeviceEntry is a description of MLC (or any beam limiting device such as symmetric or asymmetric jaws).

DICOM standard describes two kinds of multi-leaf collimators: "MLCX" - leaves moves along X-axis, "MCLY" - leaves moves along Y-axis.

LeafJawPositionsMLC[0] - leaf position vector for MLCX
LeafJawPositionsMLC[1] - leaf position vector for MLCY

Leaf position vector stores data as it had been written with in DICOM file. (60 leaf pairs = 120 leaf positions)

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: I will fix that.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: Ok.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: I will be a problem if RTPlan has more than one BeamSequence. I assume, for the first time, it is easier to test everything with only one BeamSequence.

MichaelColonel: You are right, it doesn't look good. I will rework that part.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
cpinter: Thanks for the explanation!

cpinter: Could you add some of this text to the code as comment? Adding meaningful comments facilitates maintenance and future improvements a lot! Thanks

MichaelColonel:
Some changes. Beam polydata for MLC was rewritten from scratch to use both MLC and Jaws information.

Some artifacts are still possible, when beam polydata is generated.

I don't know how to put table node correctly in hierarchy as a child of RTPlan node or beam node, and how to draw beam always foreground.

cpinter:
Thank you very much, I really appreciate your time!

We are in the middle of moving abroad, but I'll try to make some time to review it. If everything seems alright except for the two things you mention then I'll integrate it and do those myself.

Cheers!

MichaelColonel:
Beam is not always foreground is no longer an issue. Beam node doesn't update beam poly data, when i set MLC positions table node using 

` beamNode->SetAndObserveMLCPositionTableNode(tableNode); `

It updates polydata when i click on "Edit properties..." action in popup menu on a beam node. 

Basically two problems remains (IMHO):
1. Proper position of MLC table node within hierarchy, maybe better name for the MLC table node as well.
2. Beam node should observe changes in a MLC table node, and updates beam polydata
 
image

cpinter:
Perfect, thank you! Also thanks @Sunderlandkyl  for merging! Now I am gradually getting back to normal, and soon I will take a look at the two outstanding issues mentioned above.

Issue #115, opened 2019-08-08
Add support for multi-leaf collimators


cpinter:
* Implement loading of MLC beams (sample data for example the "corvus-6.2.2-phantom" dataset https://pocus.cs.queensu.ca/#collection/5cc8811601d4930406c40465/folder/5ccb15f101d4930406c4046e)
* Revive MLC code in beam node
* Allow handling MLCs in beam planning

cpinter:
@MichaelColonel This is where the MLC related work should be tracked.

So what exists in SlicerRT right now is a way to set the leaf positions
https://github.com/SlicerRt/SlicerRT/blob/master/Beams/MRML/vtkMRMLRTBeamNode.h#L101-L105
However it uses double array nodes, and I think it would be much better using table nodes (vtkMRMLTableNode). Do you know how are MLC positions stored in DICOM?

There is also some code that generates the beam model from the MLC positions, see
https://github.com/SlicerRt/SlicerRT/blob/master/Beams/MRML/vtkMRMLRTBeamNode.cxx#L483-L574
I think this should be OK, just need to change all to use table nodes, and add importing MLCs from DICOM (or whatever feature you need around MLCs)
(https://github.com/SlicerRt/SlicerRT/blob/master/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx#L637-L640)

gregsharp:
MLC positions are stored as a sequence of control points.  MLC positions may change between control points.  Each control point also specifies the gantry angle, collimator angle, jaw positions, and cumulative MUs (and several other things as well).   

The plan also contains information about the number of MLC leaves and the widths of those leaves.  However, it does not contain all needed information, for example leaf position limits.

cpinter:
Thanks, @gregsharp ! What is the format of the MLC positions? Just a number for each leaf for each control point?

gregsharp:
Correct.  A single number for each leaf position at each control point.

Regarding using a vtkMRMLTableNode, that would only be for the "Current leaf position"?  Or do you envision a sequence of vtkMRMLTableNode for multiple control points?

cpinter:
You're probably right and we should put all this in a sequence.

lassoan:
Storing in a sequence node would have the advantage that you would get browsing between time points for free. You could add any other nodes (markup fiducials, transforms, etc) to change in sync. 

[...]

PR #124, opened 2019-12-30
ENH: Initial RTIonPlan and multiple control points support

MichaelColonel:
Greetings,

vtkSlicerDicomRtReader class - RTIonPlan support and control points support, some C-like arrays replaced with std::array for easier copy and assign operations. Scan spot position map inital support for modulated scan type of rt ion beam. Fraction module from DICOM is used to check if beams are available.

vtkSlicerDicomRtImportExportModuleLogic class - each control point is represented as a separate beam. Table nodes are set a children of a corresponding beam nodes.

vtkMRMLRTIonBeamNode class is added for ion beam parameters and visualization.

vtkMRMLRTBeamNode multiple visible sections of MLC opening have been added (corvus-6.2.2-phantom example data control point 10 & 11)

I can split this pull request on several parts if its very big.

### DIFF
--- a/Beams/MRML/vtkMRMLRTBeamNode.cxx
+++ b/Beams/MRML/vtkMRMLRTBeamNode.cxx
@@ -47,6 +47,7 @@
 #include <vtkTable.h>
 #include <vtkCellArray.h>
 #include <vtkAppendPolyData.h>
+#include <vtkCubeSource.h>
 
 //------------------------------------------------------------------------------
 const char* vtkMRMLRTBeamNode::NEW_BEAM_NODE_NAME_PREFIX = "NewBeam_";
@@ -62,148 +63,7 @@ static const char* CONTOUR_BEV_REFERENCE_ROLE = "contourBEVRef";
 namespace
 {
 
-typedef std::vector< std::pair< double, double > > PointVector;
-typedef std::vector< std::array< double, 4 > > LeafDataVector;
-typedef std::vector< std::pair< LeafDataVector::iterator, LeafDataVector::iterator > > SectionVector;
-typedef std::array< double, 6 > JawsParameters;
-typedef std::pair< bool, bool > MLCType;
-
 bool(*AreEqual)( double, double) = vtkSlicerRtCommon::AreEqualWithTolerance;
-
-void
-CreateMLCPointsFromSectionBorders( const JawsParameters& jawsParameters,
-  const MLCType& mlcType, const SectionVector::value_type& sectionBorder, 
-  PointVector& side12)
-{
-  const double& jawBegin = jawsParameters[0];
-  const double& jawEnd = jawsParameters[1];
-  const double& X1Jaw = jawsParameters[2];
-  const double& X2Jaw = jawsParameters[3];
-  const double& Y1Jaw = jawsParameters[4];
-  const double& Y2Jaw = jawsParameters[5];
-
-  bool typeMLCX = mlcType.first;
-  bool typeMLCY = mlcType.second;
-
-  PointVector side1, side2; // temporary vectors to save visible points
-
-  LeafDataVector::iterator firstLeafIterator = sectionBorder.first;
-  LeafDataVector::iterator lastLeafIterator = sectionBorder.second;
-  LeafDataVector::iterator firstLeafIteratorJaws = firstLeafIterator;
-  LeafDataVector::iterator lastLeafIteratorJaws = lastLeafIterator;
-
-  // find first and last visible leaves using Jaws data
-  for ( auto it = firstLeafIterator; it <= lastLeafIterator; ++it)
-  {
-    double& bound1 = (*it)[0]; // leaf begin boundary
-    double& bound2 = (*it)[1]; // leaf end boundary
-    if (bound1 <= jawBegin && bound2 > jawBegin)
-    {
-      firstLeafIteratorJaws = it;
-    }
-    else if (bound1 <= jawEnd && bound2 > jawEnd)
-    {
-      lastLeafIteratorJaws = it;
-    }
-  }
-
-  // find opened MLC leaves into Jaws opening (logical AND)
-  if (firstLeafIteratorJaws != firstLeafIterator)
-  {
-    firstLeafIterator = std::max( firstLeafIteratorJaws, firstLeafIterator);
-  }
-  if (lastLeafIteratorJaws != lastLeafIterator)
-  {
-    lastLeafIterator = std::min( lastLeafIteratorJaws, lastLeafIterator);
-  }
-
-  // add points for the visible leaves of side "1" and "2"
-  // into side1 and side2 points vectors
-  for ( auto it = firstLeafIterator; it <= lastLeafIterator; ++it)
-  {
-    double& bound1 = (*it)[0]; // leaf begin boundary
-    double& bound2 = (*it)[1]; // leaf end boundary
-    double& pos1 = (*it)[2]; // leaf position "1"
-    double& pos2 = (*it)[3]; // leaf position "2"
-    if (typeMLCX)
-    {
-      side1.push_back({ std::max( pos1, X1Jaw), bound1});
-      side1.push_back({ std::max( pos1, X1Jaw), bound2});
-      side2.push_back({ std::min( pos2, X2Jaw), bound1});
-      side2.push_back({ std::min( pos2, X2Jaw), bound2});
-    }
-    else if (typeMLCY)
-    {
-      side1.push_back({ bound1, std::max( pos1, Y1Jaw)});
-      side1.push_back({ bound2, std::max( pos1, Y1Jaw)});
-      side2.push_back({ bound1, std::min( pos2, Y2Jaw)});
-      side2.push_back({ bound2, std::min( pos2, Y2Jaw)});
-    }
-  }
-
-  // intersection between Jaws and MLC boundary (logical AND) lambda
-  auto intersectJawsMLC = [ jawBegin, jawEnd, typeMLCX, typeMLCY](PointVector::value_type& point)
-  {
-    double& leafBoundary = point.second;
-    if (typeMLCX) // JawsY and MLCX
-    {
-      leafBoundary = point.second;
-    }
-    else if (typeMLCY) // JawsX and MLCY
-    {
-      leafBoundary = point.first;
-    }
-
-    if (leafBoundary <= jawBegin)
-    {
-      leafBoundary = jawBegin;
-    }
-    else if (leafBoundary >= jawEnd)
-    {
-      leafBoundary = jawEnd;
-    }
-  };
-
-  // apply lambda to side "1"
-  std::for_each( side1.begin(), side1.end(), intersectJawsMLC);
-  // apply lambda to side "2"
-  std::for_each( side2.begin(), side2.end(), intersectJawsMLC);
-
-  // reverse side "2"
-  std::reverse( side2.begin(), side2.end());
-
-  // fill real points vector side12 without excessive points from side1 vector
-  PointVector::value_type& p = side1.front(); // start point
-  double& px = p.first; // x coordinate of p point
-  double& py = p.second; // y coordinate of p point
-  side12.push_back(p);
-  for ( size_t i = 1; i < side1.size() - 1; ++i)
-  {
-    double& pxNext = side1[i + 1].first; // x coordinate of next point
-    double& pyNext = side1[i + 1].second; // y coordinate of next point
-    if (!AreEqual( px, pxNext) && !AreEqual( py, pyNext))
-    {
-      p = side1[i];
-      side12.push_back(p);
-    }
-  }
-  side12.push_back(side1.back()); // end point
-
-  // same for the side2 vector
-  p = side2.front();
-  side12.push_back(p);
-  for ( size_t i = 1; i < side2.size() - 1; ++i)
-  {
-    double& pxNext = side2[i + 1].first;
-    double& pyNext = side2[i + 1].second;
-    if (!AreEqual( px, pxNext) && !AreEqual( py, pyNext))
-    {
-      p = side2[i];
-      side12.push_back(p);
-    }
-  }
-  side12.push_back(side2.back());
-}
 
 } // namespace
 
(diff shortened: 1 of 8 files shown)


Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -57,6 +58,155 @@ static const char* MLCPOSITION_REFERENCE_ROLE = "MLCPositionRef";
 static const char* DRR_REFERENCE_ROLE = "DRRRef";
 static const char* CONTOUR_BEV_REFERENCE_ROLE = "contourBEVRef";
 
+//------------------------------------------------------------------------------
+namespace
+{
+
+typedef std::vector< std::pair< double, double > > PointVector;
```
cpinter: Please rename variable to something more meaningful (points of what?)

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -57,6 +58,155 @@ static const char* MLCPOSITION_REFERENCE_ROLE = "MLCPositionRef";
 static const char* DRR_REFERENCE_ROLE = "DRRRef";
 static const char* CONTOUR_BEV_REFERENCE_ROLE = "contourBEVRef";
 
+//------------------------------------------------------------------------------
+namespace
+{
+
+typedef std::vector< std::pair< double, double > > PointVector;
+typedef std::vector< std::array< double, 4 > > LeafDataVector;
```
cpinter: Same here, please don't use the word data, but call it something more informative.

cpinter: Same here, please don't use the word data, but call it something more informative.

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -57,6 +58,155 @@ static const char* MLCPOSITION_REFERENCE_ROLE = "MLCPositionRef";
 static const char* DRR_REFERENCE_ROLE = "DRRRef";
 static const char* CONTOUR_BEV_REFERENCE_ROLE = "contourBEVRef";
 
+//------------------------------------------------------------------------------
+namespace
```
cpinter, line 63: Why do you use a namespace here? It would be much better making this part of the class

Beams/MRML/vtkMRMLRTIonBeamNode.cxx
```diff
@@ -0,0 +1,514 @@
...
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLRTIonBeamNode::~vtkMRMLRTIonBeamNode()
+{
+  this->SetBeamDescription(nullptr);
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLRTIonBeamNode::WriteXML(ostream& of, int nIndent)
```
cpinter, line 87: Please use the new macros defined here
https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLNodePropertyMacros.h
See examples here
https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLDisplayNode.cxx#L110
Also for ReadXMLAttributes, Copy, and PrintSelf

Beams/MRML/vtkMRMLRTIonBeamNode.cxx
```diff
@@ -0,0 +1,514 @@
...
+  Superclass::CreateNewBeamTransformNode();
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLRTIonBeamNode::SetVSAD( double xComponent, double yComponent)
+{
+  this->VSADx = xComponent;
+  this->VSADy = yComponent;
+  this->Modified();
+  this->InvokeCustomModifiedEvent(vtkMRMLRTBeamNode::BeamGeometryModified);
```
cpinter, line 242: In some places you use Superclass, in others the class name itself. Please make it consistent.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -77,18 +77,53 @@ class vtkSlicerDicomRtReader::vtkInternal
   public:
     BeamLimitingDeviceEntry()
       :
-      Distance(-1.),
-      NumberOfLeafJawPairs(0)
+      SourceIsoDistance(400.),
+      NumberOfPairs(0)
```
cpinter: Why did you rename this variable to have a less informative name? I'm actually curious, so if it's justifiable then no need to change, but in general we prefer longer but unambiguoug names. Not every reader of the code is an expert in every aspect of RT.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -77,18 +77,53 @@ class vtkSlicerDicomRtReader::vtkInternal
...
+    }
+    int Number;
+    std::string ID;
+    std::string AccessoryCode;
+    std::string Type; // "ANALOG", "BINARY"
+    std::string Description;
+  };
+
+  /// Structure storing a treatment compensator parameters associated with beam
+  class CompensatorEntry
```
cpinter, line 112: Is this class and BlockEntry future work? Since they don't contain anything... In that case please add "//TODO: ..." comments

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -103,7 +138,7 @@ class vtkSlicerDicomRtReader::vtkInternal
     unsigned int Number;
     std::string Name;
     std::string Description;
-    double DisplayColor[3];
+    std::array< double, 3 > DisplayColor;
```
cpinter, line 144: I believe we used a static array because of VTK set/get macros. Is there a good reason to change this?

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -182,7 +322,7 @@ class vtkSlicerDicomRtReader::vtkInternal
   public:
     ChannelEntry()
     {
-      Number = -1;
+      Number = 0;
```
cpinter: Please explain why you changed this. -1 is an obvious "uninitialized" state, while 0 is valid, so this changes the way a newly instantiated object can be handled.

cpinter (changes requested):
Happy new year :) Thanks again for the great work! I left a few comments.

Please also
- Squash these 47 commits into a single one (or 2-3 commits if they can be cleanly separated) and add descriptive comments in it, explaining in detail what changed and why (for example a line about each original commit, since you made them separately on purpose)
- Run the automated tests and make sure no new failures happen

Thank you!

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: There is only one method which use DisplayColor
`double* vtkSlicerDicomRtReader::GetRoiDisplayColor(unsigned int internalIndex)`.
The return value is a pointer, so no difference. Everything else is a DCMTK specific.

No setters to change a display color, within internal elements of the reader.

I that case i had thought it's better to use std::array, rather C array (no copy, assign problems).

I can roll it back, if its required to use C array.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: I will return previous names.

Beams/MRML/vtkMRMLRTIonBeamNode.cxx
MichaelColonel: It will be Superclass only for overridden methods and inherited constructor.

Beams/MRML/vtkMRMLRTBeamNode.cxx
MichaelColonel: By means of vtkInternal, or just private?

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: I will do this.

Beams/MRML/vtkMRMLRTIonBeamNode.cxx
MichaelColonel: I fixed that, not so sure about Copy method though.

MichaelColonel:
Squashed to 6 commits.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
cpinter: If it's only internal then it's fine like this, thanks!

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
cpinter: Thanks! You also changed some names to be more informative, so those should be kept (like the SourceIsoDistance)

Beams/MRML/vtkMRMLRTBeamNode.cxx
cpinter: If there is no internal class yet, then no need to bother, just private. Otherwise internal please.

cpinter:
Thank you very much! I'll take a look shortly and integrate.
Please let me know about the automated tests.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: SourceIsoDistance parameter has different physical meaning for RTPlan and RTIonPlan but the same DICOM tag (300A,00BA) 

1. For RTPlan its "Source to Beam Limiting Device Distance";
2. For RTIonPlan its "Isocenter to Beam Limiting Device Distance".

In any case the name of the parameters isn't obvious, but i will keep it.

MichaelColonel:
23 failed out of 70

	 12 - py_DicomRtImportTest (Failed)
	 18 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseProstate_Base (Failed)
	 19 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseProstate_CERR (Failed)
	 20 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseEnt_CERR (Failed)
	 21 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseProstate_Eclipse (Failed)
	 22 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseEnt_Eclipse (Failed)
	 23 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseProstate_CERR_AutomaticOversampling (Failed)
	 24 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseEnt_CERR_AutomaticOversampling (Failed)
	 25 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseProstate_Eclipse_AutomaticOversampling (Failed)
	 26 - vtkSlicerDoseVolumeHistogramModuleLogicTest_EclipseEnt_Eclipse_AutomaticOversampling (Failed)
	 27 - vtkSlicerDoseVolumeHistogramModuleLogicTest_DoseSurfaceHistogram_EclipseEnt_Base_Inside (Failed)
	 28 - vtkSlicerDoseVolumeHistogramModuleLogicTest_DoseSurfaceHistogram_EclipseEnt_Base_Outside (Failed)
	 29 - vtkSlicerDoseVolumeHistogramModuleLogicTest_DoseSurfaceHistogram_EclipseProstate_Base_Inside (Failed)
	 30 - vtkSlicerDoseVolumeHistogramModuleLogicTest_DoseSurfaceHistogram_EclipseProstate_Base_Outside (Failed)
	 37 - vtkSlicerSegmentMorphologyModuleLogicTest_EclipseProstate_Expand (Failed)
	 38 - vtkSlicerSegmentMorphologyModuleLogicTest_EclipseProstate_Shrink (Failed)
	 39 - vtkSlicerSegmentMorphologyModuleLogicTest_EclipseProstate_Union (Failed)
	 40 - vtkSlicerSegmentMorphologyModuleLogicTest_EclipseProstate_Intersect (Failed)
	 41 - vtkSlicerSegmentMorphologyModuleLogicTest_EclipseProstate_Subtract (Failed)
	 42 - vtkSlicerSegmentMorphologyModuleLogicTest_EclipseProstate_Intersect_ApplyTransform (Failed)
	 45 - vtkSlicerSegmentComparisonModuleLogicTest_EclipseProstate_Base (Failed)
	 46 - vtkSlicerSegmentComparisonModuleLogicTest_EclipseProstate_SameInput (Failed)
	 47 - vtkSlicerSegmentComparisonModuleLogicTest_EclipseProstate_Transformed (Failed)

MichaelColonel:
Offtopic question. If i have been needed a MLC visualization like in XiO, should i create a custom node for MLC model, or i could use a beam node?

From this [post](https://github.com/SlicerRt/SlicerRT/issues/115#issuecomment-559332125)

cpinter:
> 23 failed out of 70

Well, those will need to be fixed. I see that these failures are the same as what SlicerRT has had since the end of October. It would be great if the tests could be fixed by the developer team so that you can (after a rebase) see if your changes have broken anything or not.
I added a ticket for this (https://github.com/SlicerRt/SlicerRT/issues/125) and contacted the team about it. Please stand by for a bit until we see if you can rebase and try it again soon.

cpinter:
About the XiO question: Is this visualization in the 3D view? If so, then special overlay would be needed to achive the same, which will require substantial work. However, you may not need it in the 3D view, in which case it is a simple slice intersection, which could already work with your changes.
Thoughts @gregsharp @lassoan ?

MichaelColonel:
The picture (part of it) is taken from [SlicerRtData](https://github.com/SlicerRt/SlicerRtData/blob/master/xio-4.64-phantom-prostate-beams-mlc/_MLC.jpg).

lassoan:
> About the XiO question: Is this visualization in the 3D view? If so, then special overlay would be needed to achive the same, which will require substantial work. However, you may not need it in the 3D view, in which case it is a simple slice intersection, which could already work with your changes.
Thoughts @gregsharp @lassoan ?

Since it is a beam's eye view, the camera position, orientation, and viewing angle has match the beam parameters and cannot be rotated. Since the camera cannot be rotated and viewing angle cannot be changed, it does not matter if you display the grid as a 2D or 3D actor. In 3D views it is very easy to display 3D (using model nodes), so I would just go with that. If you create a model that contains projection of the collimator leaves (their shadow) then you can use the exact same model in 2D views (as slice intersection) and in 3D view. Another advantage of this approach is that if for some reason the camera gets rotated or any way misaligned with the beam's eye view, it will still be correct (but you'll see 3D beams and not just a 2D grid). In beam's eye 3D views you can choose to disable camera rotation similarly to how it is done for slice views [here](https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository#Disable_certain_user_interactions_in_slice_views).

cpinter:
@MichaelColonel The problem that made most of the tests fail has been fixed by @Sunderlandkyl, and the dashboard shows only one failing test. So I think if you can build this branch against the latest Slicer and only the same tests fails (py_DicomRtImportTest, which tests importing/loading DICOM but not ion plan obviously - it may be added to it though), then I'll integrate this. Thanks again!

MichaelColonel:
All tests have passed except py_DicomRtImportTest. It failed because number of loaded beam nodes, and total number of nodes in hierarchy had increased (each control point is a separated beam node).

After some minor changes:

1.   `self.assertEqual( len( slicer.util.getNodes('vtkMRMLModelHierarchyNode*') ), 11 )`
2.   `self.assertEqual( shNode.GetNumberOfItems(), 42 )`

it also passed.

These changes are not in pull request.

cpinter:
@MichaelColonel you are great :) Now that the tests pass, I'll integrate this change, but not before you give me the green light, because I have seen pushes recently, and not sure if you want to add more improvements.

I'll make the changes in py_DicomRtImportTest and close the related ticket. Thanks again!

MichaelColonel:
The pull request is ready to be merged, latest pushes were minor API style changes - GetBeamControlPoint* instead of GetControlPoint*.

PR #132, opened 2020-01-24
BUG: MLC beam polydata fix when a leaf pair is outside the jaws

MichaelColonel:
Data to test
https://github.com/SlicerRt/SlicerRtData/tree/master/xio-4.64-phantom-prostate-beams-imrt


cpinter:
Excellent, thank you!

PR #134, opened 2020-03-17
Fix implicit instantiation of std::array for Windows and MacOS build

MichaelColonel:


cpinter:
Could you please explain in the commit message the reason and purpose of this change?

MichaelColonel:
There are several errors for [Windows](http://slicer.cdash.org/viewBuildError.php?buildid=1857171) and [MacOS](http://slicer.cdash.org/viewBuildError.php?buildid=1857245) builds. [Linux](http://slicer.cdash.org/viewBuildError.php?buildid=1856968) build doesn't have such problem, since some c++11 headers are included implicitly. I can only test linux distribution at home.

cpinter:
Then please add this into the commit message, and change the prefix to "COMP:"

MichaelColonel:
Could you check that commit name is correct now?

cpinter:
I don't mean to be nitpicking here, but the commit message is very generic. You already described it in greater detail in the previous commit message and then here in a comment. Could you just copy-paste it there? It is not advantageous trying to be as concise as possible in the commit messages, but the opposite! Thank you!

MichaelColonel:
I have added a more detailed commit message.

cpinter:
Excellent, thank you very much!

PR #141, opened 2020-04-22
Update BatchStructureSetConversion.py

codecrazer:
change the name of saved file to patient ID

### DIFF
--- a/BatchProcessing/BatchStructureSetConversion.py
+++ b/BatchProcessing/BatchStructureSetConversion.py
@@ -321,16 +321,20 @@ def main(argv):
     if exist_db:
       logging.info('BatchStructureSet running in existing database mode')
       DICOMUtils.openDatabase(input_folder)
-      all_patients = slicer.dicomDatabase.patients()
+      db = slicer.dicomDatabase
+      all_patients = db.patients()
       logging.info('Must Process Patients %s' % len(all_patients))
-
       for patient in all_patients:
-        slicer.mrmlScene.Clear(0) # clear the scene
-        DICOMUtils.loadPatientByUID(patient)
-        output_dir = os.path.join(output_folder,patient)
+        slicer.mrmlScene.Clear(0) # clear the scen
+        studyList=db.studiesForPatient(patient) # code modified from https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository 
+        seriesList=db.seriesForStudy(studyList[0])
+        fileList=db.filesForSeries(seriesList[0])
+        print(db.fileValue(fileList[0],'0010,0020'))
+        DICOMUtils.loadPatientByUID(patient)       
+        output_dir = os.path.join(output_folder,db.fileValue(fileList[0],'0010,0020'))
         if not os.access(output_dir, os.F_OK):
           os.mkdir(output_dir)
-        save_rtslices(output_dir, use_ref_image)
+        save_rtslices(output_dir)
 
     else:
       ref_image_node_id = None


BatchProcessing/BatchStructureSetConversion.py
```diff
@@ -321,16 +321,20 @@ def save_rtslices(output_dir, use_ref_image, ref_image_node_id=None):
...
+        slicer.mrmlScene.Clear(0) # clear the scen
+        studyList=db.studiesForPatient(patient) # code modified from https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository 
+        seriesList=db.seriesForStudy(studyList[0])
+        fileList=db.filesForSeries(seriesList[0])
+        print(db.fileValue(fileList[0],'0010,0020'))
+        DICOMUtils.loadPatientByUID(patient)       
+        output_dir = os.path.join(output_folder,db.fileValue(fileList[0],'0010,0020'))
         if not os.access(output_dir, os.F_OK):
           os.mkdir(output_dir)
-        save_rtslices(output_dir, use_ref_image)
```
cpinter, line 333: Please add the argument back, this would break the reference image feature.

BatchProcessing/BatchStructureSetConversion.py
```diff
@@ -321,16 +321,20 @@ def save_rtslices(output_dir, use_ref_image, ref_image_node_id=None):
...
-      all_patients = slicer.dicomDatabase.patients()
+      db = slicer.dicomDatabase
+      all_patients = db.patients()
       logging.info('Must Process Patients %s' % len(all_patients))
-
       for patient in all_patients:
-        slicer.mrmlScene.Clear(0) # clear the scene
-        DICOMUtils.loadPatientByUID(patient)
-        output_dir = os.path.join(output_folder,patient)
+        slicer.mrmlScene.Clear(0) # clear the scen
```
cpinter, line 328: You introduced a typo here

BatchProcessing/BatchStructureSetConversion.py
```diff
@@ -321,16 +321,20 @@ def save_rtslices(output_dir, use_ref_image, ref_image_node_id=None):
...
+        studyList=db.studiesForPatient(patient) # code modified from https://www.slicer.org/wiki/Documentation/Nightly/ScriptRepository
```
cpinter, line 329: This is more complicated than necessary. I think this would work better:
`slicer.dicomDatabase.fieldForPatient('PatientID', patient)`

In addition, I think the database UID should be kept, because in DICOM the patient ID is not necessarily unique, and it is unique together with the patient name, but it is fragile to use patient name in path. With the database patient UID though, it is unique within the Slicer database. Something like this:
```
patientID = slicer.dicomDatabase.fieldForPatient('PatientID', patient)
folderName = str(patient) + '_' + patientID
```
@lassoan What do you think?

cpinter (changes requested):
Thanks for the PR! I added some suggestions.

BatchProcessing/BatchStructureSetConversion.py
lassoan: Patient ID can be really anything: it is often set to empty during image anonymization (even though it is not valid), can be too long to be used as filename, and may contain special characters that cannot be used in filenames. So, they are definitely much more risky to use than the database's patient UID.

I would recommend to make addint patient name+id to folder name optional and sanitize the string (remove special characters, limit maximum size, etc.).

BatchProcessing/BatchStructureSetConversion.py
codecrazer: What's the difference between "save_rtslices(output_dir, use_ref_image)" and "save_rtslices(output_dir)". I know this is a previous issue, but after reading the post, I can't realize what is reference image.

cpinter: If you specify a reference image, then the saved NRRDs will have that dimension instead of the segments being cropped to their "effective extent". But even if you don't understand a feature, removing it is not a good idea :)

cpinter:
The PR is four years old and there has been no answer to the requests, closing.

PR #144, opened 2020-05-09
STYLE: More robust check and detailed debug information for MLC table node

MichaelColonel:
Tests are OK, except python ones.


cpinter:
Thank you!

Issue #145, opened 2020-05-20
RT plan loading issues


cpinter:
It takes around 10 minutes to load the Eclipse ENT test dataset. It only happens if I load the RT plan as well. It seems an unreasonably long time even if the plan is complex. In addition, it created dozens of beam nodes, with one MLC table in each, see below.

@MichaelColonel
1. Do you have an idea why it takes so long?
2. Is it a feature that so many beam nodes are created? By design we should have as many beam nodes as beams

![image](https://user-images.githubusercontent.com/1325980/82458642-d793dc00-9aae-11ea-8aba-cccbacf56629.png)

MichaelColonel:
1. It definitely shouldn't load data so long, i will try to fix that issue ASAP.
2. If beam is intensity modulated, or some king of conformal in plan, then each beam can have multiple (dozens or hundreds) control points, and each control point is loaded as a separated beam. That is not a best idea, but there is nothing better.

The main reason why it take so much time, is the ImportExportLogic try to update (redraw) each beam when a new beam was added into the scene.

cpinter:
Thank you for the quick answer! I appreciate you taking a look.

You're right, we did discuss the control points. Maybe they should be grouped to folders or somehow made easier to manage, because this very long list is not useful at all unfortunately. @gregsharp do you have any idea for this?

lassoan:
It sounds like the most appropriate would be to store these control points in Sequence nodes. You can store thousands of nodes in one sequence node with small overhead. You can browse between sequence items using a slider or even replay them. You can expose one or more items from a sequence, and of course you have full access to all the nodes from code.

Sequences are now part of Slicer core (in Slicer-4.11) before that it was a separate extension.

MichaelColonel:
As a temporary fix i can disable loading of multiple control points, so it will be only one control point per beam, as it was previously.

If batch processing is activated the loading is fast, but for the each beam one must manually invoke BeamTransformModified and BeamGeometryModified events (using Beams module) for proper visualization. For some reason BeamTransformModified event takes quite amount of time.

cpinter:
Thanks @lassoan and @MichaelColonel! Let me take a closer look at the options before any quick fixes are made.

cpinter:
In the latest version the loading time is similar to what we had before, and with the [sequences change](https://github.com/SlicerRt/SlicerRT/pull/147) the beam nodes are not overwhelming with the one node per control point, but there is one item for each beam.

PR #146, opened 2020-05-20
BUG: Multiple control points loading is disabled, only control point 0 will be loaded (issue #145)

MichaelColonel:
Temporary fix for issue #145. Tests are passed, except python scripts.

If multiple control points are not required immediately.

### DIFF
--- a/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
+++ b/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
@@ -805,7 +805,12 @@ bool vtkSlicerDicomRtImportExportModuleLogic::vtkInternal::LoadExternalBeamPlan(
         nofCointrolPoints = 1; // will be used only control point 0
       }
     }
-    
+
+    //TODO: Implement correct loading of multiple control points (e.g. sequence nodes)
+    singleBeam = true;
+    vtkWarningWithObjectMacro( this->External, "LoadExternalBeamPlan: Multiple control points loading are currently disabled!");
+    nofCointrolPoints = 1; // will be used only control point 0
+
     for ( unsigned int cointrolPointIndex = 0; cointrolPointIndex < nofCointrolPoints; ++cointrolPointIndex)
     {
       // Create the beam node for each control point


lassoan:
This workaround could be good for Slicer-4.10, but for Slicer-4.11 it should be very easy to create a sequence node instead. You create the nodes the same way as before, but instead of adding it to the main scene, you add them to a vtkMRMLSequenceNode using [SetDataNodeAtValue](https://github.com/Slicer/Slicer/blob/master/Libs/MRML/Core/vtkMRMLSequenceNode.h#L104). Then you add the sequence node to the main scene. For user's convenience, you can create a sequence browser node. I would be happy help with any specific questions that you might have.

MichaelColonel:
Agreed. I will try to add sequence nodes instead.

cpinter:
If you can take a stab at it @MichaelColonel we appreciate it! Let us know if you have any questions.

MichaelColonel:
First round of questions:
1. Before beam node will be added to the sequence node, it must be properly loaded, transformed, etc. Is it possible to invoke custom event (or events) for beam node, while these beam nodes are played by a sequence browser node? It is required, because beam transformation `beamNode->InvokeCustomModifiedEvent(vtkMRMLRTBeamNode::BeamTransformModified);` becomes super laggy, then hundreds of such nodes are in the plan node.

2. Should a sequence node be added to the plan node, so the user will be able to see what king of beam is it (static single beam or dynamic beam with multiple control points)?

3. How the sequence node must be added to the subject hierarchy?

MichaelColonel:
Another question:

Does the [subject hierarchy plugin](https://github.com/SlicerRt/SlicerRT/blob/master/Beams/SubjectHierarchyPlugins/qSlicerSubjectHierarchyRTBeamPlugin.h) required in order to make sequence node visible in subject hierarchy node?

lassoan:
> Before beam node will be added to the sequence node, it must be properly loaded, transformed, etc. Is it possible to invoke custom event (or events) for beam node, while these beam nodes are played by a sequence browser node? It is required, because beam transformation beamNode-

Loading: you load into a data node the same way. The only difference is that the data node is in the sequence node's internal scene.

Transform: you apply the transform to the proxy node (the node that represents a selected sequence item in the main scene). If the transform is time-varying, too, then the transform node in the main scene is a proxy node, too.

>InvokeCustomModifiedEvent(vtkMRMLRTBeamNode::BeamTransformModified); becomes super laggy, then hundreds of such nodes are in the plan node.

There won't be hundreds of nodes in the main scene. All the time-varying nodes will be in internal scenes in sequence nodes.

> Should a sequence node be added to the plan node, so the user will be able to see what king of beam is it (static single beam or dynamic beam with multiple control points)?

You can retrieve sequence node from the proxy node, but it probably make sense to keep a node reference directly to the beam sequence node.

> How the sequence node must be added to the subject hierarchy?

Usually we don't add subject hierarchy plugin for the sequence node: the user selects a time point with the sequence browser slider and works with proxy nodes corresponding to that time point. However, if you want the beam sequence to show up (and not just the current beam) then you can certainly add a subject hierarchy plugin for it.

> Does the subject hierarchy plugin required in order to make sequence node visible in subject hierarchy node?

Yes.

MichaelColonel:
Last question for today:

Can a virtual output nodes created by a sequence browser, in my simple case beam node and transformation node from corresponding sequences, be manipulated as a usual nodes, for example added to the plan?

lassoan:
Yes, you can use proxy nodes (old term: virtual output nodes) the same way as any other regular nodes.

PR #154, opened 2020-09-12
ENH: Allow loading invalid deformable DICOM SRO created by ImSimQA software

lassoan:
This workaround should be removed once ImSimQA updates its software.

See discussion here: https://discourse.slicer.org/t/application-of-deformation-field-not-doing-anything/13402/7


lassoan:
I'll merge this now. We can remove it later to keep the clone cleaner and more strictly enforce DICOM standard conformance.

PR #155, opened 2020-09-21
BUG: Fix shared labelmap bugs in SegmentComparison/SegmentMorphology

Sunderlandkyl:
In SegmentComparison/SegmentMorphology, the vtkSlicerSegmentationsModuleLogic::GetSegmentBinaryLabelmapRepresentation function was used to get the binary labelmaps. This function returns the internal "shared" representation of the segments, rather than the individual segment labelmap. This caused issues in SegmentComparison/SegmentMorphology, which expected only a single segment in each labelmap.

Fixed by changing vtkSlicerSegmentationsModuleLogic::GetSegmentBinaryLabelmapRepresentation to vtkSegmentationNode::GetBinaryLabelmapRepresentation, which retrieves the individual non-shared labelmap for each representation.

### DIFF
--- a/SegmentComparison/Logic/vtkSlicerSegmentComparisonModuleLogic.cxx
+++ b/SegmentComparison/Logic/vtkSlicerSegmentComparisonModuleLogic.cxx
@@ -130,16 +130,14 @@ std::string vtkSlicerSegmentComparisonModuleLogicPrivate::GetInputSegmentsAsPlmV
 
   // Get segment binary labelmaps
   vtkSmartPointer<vtkOrientedImageData> referenceSegmentLabelmap = vtkSmartPointer<vtkOrientedImageData>::New();
-  if ( !vtkSlicerSegmentationsModuleLogic::GetSegmentBinaryLabelmapRepresentation(
-    referenceSegmentationNode, referenceSegmentID, referenceSegmentLabelmap ) )
+  if (!referenceSegmentationNode->GetBinaryLabelmapRepresentation(referenceSegmentID, referenceSegmentLabelmap))
   {
     std::string errorMessage("Failed to get binary labelmap from reference segment: " + std::string(referenceSegmentID));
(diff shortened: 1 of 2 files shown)


SegmentComparison/Logic/vtkSlicerSegmentComparisonModuleLogic.cxx
```diff
@@ -130,21 +130,10 @@ std::string vtkSlicerSegmentComparisonModuleLogicPrivate::GetInputSegmentsAsPlmV
...
+
   vtkSmartPointer<vtkOrientedImageData> compareSegmentLabelmap = vtkSmartPointer<vtkOrientedImageData>::New();
-  if ( !vtkSlicerSegmentationsModuleLogic::GetSegmentBinaryLabelmapRepresentation(
-    compareSegmentationNode, compareSegmentID, compareSegmentLabelmap ) )
-  {
-    std::string errorMessage("Failed to get binary labelmap from reference segment: " + std::string(compareSegmentID));
-    vtkErrorMacro("GetInputSegmentsAsPlmVolumes: " << errorMessage);
-    return errorMessage;
-  }
+  referenceSegmentationNode->GetBinaryLabelmapRepresentation(compareSegmentID, compareSegmentLabelmap);
```
lassoan: We should probably still have some error check and reporting. Maybe the segment is empty or labelmap representation creation fails (out of memory, ...) and in this case it is nicer if some error message is displayed instead of showing some invalid outputs.

lassoan (commented):
Thank you, it mostly looks good, I just had one comment.

SegmentComparison/Logic/vtkSlicerSegmentComparisonModuleLogic.cxx
Sunderlandkyl: OK, added back some error checking. Now requires this PR in Slicer: https://github.com/Slicer/Slicer/pull/5193.

lassoan (approved):
Looks good, thank you!

PR #156, opened 2020-09-21
ENH: Remove SegmentMorphology extension

Sunderlandkyl:
Removing the SegmentMorphology extension, since the functionality is already present in the SegmentEditor "Margin" and "Logical operator" effects.
See discussion here: https://discourse.slicer.org/t/dicecomputation-module-does-not-calculate/1757


lassoan (approved):
Looks good, thank you!

Issue #164, opened 2020-11-05
3DSlicer factory unable to compile plastimatch because of DRR commit.


MichaelColonel:
Windows build: http://slicer.cdash.org/buildSummary.php?buildid=2053208
Linux build: http://slicer.cdash.org/buildSummary.php?buildid=2052960

It looks like making a branch from plastimatch master HEAD was a bad idea. There is an error because of some DCMTK item tag.

Solution:
1. Take previous plastimatch tag which works: 2e729bc2c7c4eaa51184ed0acb23d91e615a7ff8
2. Add needed DRR elements for SlicerRT

Here is a new branch that should compile: https://gitlab.com/MichaelColonel/plastimatch/-/commits/drr-factory-comp

cpinter:
What happened? You opened this issue, then created a PR, which was not integrated, then you closed this issue. In such a case explanation would be welcome.

MichaelColonel:
Apologies, miss click. 
There are more than one commit in that branch, so should that branch be deleted or squashed?

cpinter:
Absolutely, squash it please. I don't see a good reason for keeping five commits for one changed line. Thanks!

MichaelColonel:
PR is squashed, let's hope there will be no compilation problems on site.

Could it be possible that factory build and SuperBuild 3DSlicer has different DCMTK versions, because i don't have plastimatch compilation errors because of DCMTK (on my PC)?

UPD: DCMTK version in Plastimatch SuperBuild is 3.6.2, in 3DSlicer's DCMTK version is 3.6.5

MichaelColonel:
Factory compilation have been successful last couple of days, and i think this issue can be closed.

PR #165, opened 2020-11-05
COMP: Fix plastimatch compilation with DRR support on factory site

MichaelColonel:
Possible fix for issue #164.


cpinter:
Thanks!

Issue #96, opened 2018-12-04
Enable using percentage dose for isodose levels


cpinter:
- [ ] Enable using percentages of prescribed dose as well as absolute values
- [ ] Use the Eclipse isodose colors as default (80% deep blue, 90% lighter warmer blue, 100% yellow, 105% orange, 110% red)

(this is a new task for subtasks in the closed ticket https://github.com/SlicerRt/SlicerRT/issues/18)

Issue #36, opened 2017-09-15
Create isodose contours as percentage of max dose


cpinter:
Migrated from https://app.assembla.com/spaces/slicerrt/tickets/734-create-isodose-contours-as-percentage-of-max-dose/details

cpinter:
Closing as duplicate of #96 (its description is more detail so keeping that)

PR #168, opened 2020-11-15
WIP: ENH: Relative dose representation for isolevels

MichaelColonel:
Here is my understanding of relative dose isolevels, possible solution for #36, #96.

If RTDOSE type is "GY" then clicking on "Percentage dose isolevels" check box you can switch between absolute and relative isolevels representation. If RTDOSE type is "RELATIVE" then only relative representation is available.

Screenshot as an example:
image

### DIFF
--- a/Isodose/Logic/vtkMRMLIsodoseNode.h
+++ b/Isodose/Logic/vtkMRMLIsodoseNode.h
@@ -37,6 +37,7 @@ class vtkMRMLColorTableNode;
 class VTK_SLICER_ISODOSE_LOGIC_EXPORT vtkMRMLIsodoseNode : public vtkMRMLNode
 {
 public:
+  enum DoseUnitsType { Unknown = -1, Gy = 0, Relative = 1 };
   static const char* COLOR_TABLE_REFERENCE_ROLE;
 
   static vtkMRMLIsodoseNode *New();
@@ -94,11 +95,26 @@ public:
   vtkSetMacro(ShowDoseVolumesOnly, bool);
   vtkBooleanMacro(ShowDoseVolumesOnly, bool);
 
+  /// Get/Set reference dose value
+  vtkGetMacro(ReferenceDoseValue, double);
+  vtkSetMacro(ReferenceDoseValue, double);
+
+  /// Get/Set dose units type
+  vtkGetMacro(DoseUnits, DoseUnitsType);
+  vtkSetMacro(DoseUnits, DoseUnitsType);
+
+  /// Get/Set relative representation flag
+  vtkGetMacro(RelativeRepresentationFlag, bool);
+  vtkSetMacro(RelativeRepresentationFlag, bool);
+  vtkBooleanMacro(RelativeRepresentationFlag, bool);
+
 protected:
   vtkMRMLIsodoseNode();
   ~vtkMRMLIsodoseNode();
   vtkMRMLIsodoseNode(const vtkMRMLIsodoseNode&);
   void operator=(const vtkMRMLIsodoseNode&);
+
+  void SetDoseUnits(int doseUnits);
 
 protected:
   /// State of Show isodose lines checkbox
@@ -115,6 +131,16 @@ protected:
 
   /// State of Show dose volumes only checkbox
   bool ShowDoseVolumesOnly;
+
+  /// Type of dose units
+  DoseUnitsType DoseUnits;
+
+  /// Reference dose value
+  double ReferenceDoseValue;
+
+  /// Whether use relative isolevels representation
+  /// for absolute dose (Gy) and unknown units or not
+  bool RelativeRepresentationFlag;
 };
 
 #endif
--- a/Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
+++ b/Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
@@ -64,9 +64,11 @@
 //----------------------------------------------------------------------------
 const char* DEFAULT_ISODOSE_COLOR_TABLE_FILE_NAME = "Isodose_ColorTable.ctbl";
 const char* DEFAULT_ISODOSE_COLOR_TABLE_NODE_NAME = "Isodose_ColorTable_Default";
+const char* RELATIVE_ISODOSE_COLOR_TABLE_NODE_NAME = "Isodose_ColorTable_Relative";
 const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_MODEL_NODE_NAME_PREFIX = "IsodoseLevel_";
 const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_PARAMETER_SET_BASE_NAME_PREFIX = "IsodoseParameterSet_";
 const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_ROOT_HIERARCHY_NAME_POSTFIX = "_IsodoseSurfaces";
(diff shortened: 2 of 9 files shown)


Isodose/Logic/vtkMRMLIsodoseNode.h
```diff
@@ -37,6 +37,7 @@ class vtkMRMLColorTableNode;
 class VTK_SLICER_ISODOSE_LOGIC_EXPORT vtkMRMLIsodoseNode : public vtkMRMLNode
 {
 public:
+  enum DoseUnitsType { GY, RELATIVE, UNKNOWN_UNITS };
```
cpinter: Please see how enums are usually set up in Slicer and follow that convention

Isodose/Logic/vtkMRMLIsodoseNode.h
```diff
@@ -94,12 +95,27 @@ class VTK_SLICER_ISODOSE_LOGIC_EXPORT vtkMRMLIsodoseNode : public vtkMRMLNode
...
+  vtkSetMacro(RelativeRepresentationFlag, bool);
+  vtkBooleanMacro(RelativeRepresentationFlag, bool);
+
 protected:
   vtkMRMLIsodoseNode();
   ~vtkMRMLIsodoseNode();
   vtkMRMLIsodoseNode(const vtkMRMLIsodoseNode&);
   void operator=(const vtkMRMLIsodoseNode&);
 
+  void SetDoseUnits(int doseUnits = 0);
```
cpinter: It's strange that there is a default value for the argument in a Set function. Why?

Isodose/qSlicerIsodoseModuleWidget.cxx
```diff
@@ -217,7 +216,7 @@ void qSlicerIsodoseModuleWidget::setMRMLScene(vtkMRMLScene* scene)
     }
     else 
     {
-      vtkSmartPointer<vtkMRMLIsodoseNode> newNode = vtkSmartPointer<vtkMRMLIsodoseNode>::New();
+      vtkNew<vtkMRMLIsodoseNode> newNode;
```
cpinter, line 219: Thanks for cleaning up the code as you go.

Isodose/qSlicerIsodoseModuleWidget.cxx
```diff
@@ -430,6 +440,56 @@ void qSlicerIsodoseModuleWidget::setDoseVolumeNode(vtkMRMLNode* node)
...
+      doseUnitName = shNode->GetAttributeFromItemAncestor(
+        doseShItemID, vtkSlicerRtCommon::DICOMRTIMPORT_DOSE_UNIT_NAME_ATTRIBUTE_NAME, vtkMRMLSubjectHierarchyConstants::GetDICOMLevelStudy());
+    }
+
+    double valueRange[2];
+    vtkImageData* image = paramNode->GetDoseVolumeNode()->GetImageData();
+    image->GetScalarRange(valueRange);
+    if (!doseUnitName.compare("RELATIVE"))
+    {
+      paramNode->SetDoseUnits(vtkMRMLIsodoseNode::RELATIVE);
```
cpinter: This is how you should namespace an enum (the class name instead of the enum type). Please fix the wrong ones.

Isodose/qSlicerIsodoseModuleWidget.cxx
```diff
@@ -800,26 +973,34 @@ void qSlicerIsodoseModuleWidget::updateScalarBarsFromSelectedColorTable()
...
-    doseUnitName = shNode->GetAttributeFromItemAncestor(
-      doseShItemID, vtkSlicerRtCommon::DICOMRTIMPORT_DOSE_UNIT_NAME_ATTRIBUTE_NAME, vtkMRMLSubjectHierarchyConstants::GetDICOMLevelStudy());
-  }
-  std::string scalarBarTitle("Dose");
-  if (!doseUnitName.empty())
-  {
-    scalarBarTitle += " (" + doseUnitName + ")";
+  case vtkMRMLIsodoseNode::DoseUnitsType::GY:
+    {
+      const char* str = relativeRepresentation ? " (%)" : " (Gy)";
```
cpinter: Wrong indentation. Please fix here and below

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
```diff
@@ -523,8 +666,25 @@ void vtkSlicerIsodoseModuleLogic::CreateIsodoseSurfaces(vtkMRMLIsodoseNode* para
...
+    || doseUnits == vtkMRMLIsodoseNode::DoseUnitsType::UNKNOWN_UNITS))
+  {
+    relativeFlag = true;
+  }
+  else if (doseUnits == vtkMRMLIsodoseNode::DoseUnitsType::RELATIVE)
+  {
+    relativeFlag = true;
+  }
+  std::string isodoseName = (relativeFlag) ? 
+    vtkSlicerIsodoseModuleLogic::ISODOSE_RELATIVE_ROOT_HIERARCHY_NAME_POSTFIX :
```
cpinter, line 683: Why do we need to have two different postfixes? Can we use both at the same time?

cpinter (changes requested):
Thanks a lot, makes sense to me.

In addition to my comments in the code I find it strange that you have a collpasible button on the UI for the Prescribed or reference dose, but then below you have the levels table. It is quite confusing what belongs where.

Isodose/Logic/vtkMRMLIsodoseNode.h
MichaelColonel: > In addition to my comments in the code I find it strange that you have a collpasible button on the UI for the Prescribed or reference dose, but then below you have the levels table. It is quite confusing what belongs where.

Not quite sure what is the problem with levels table? Do you want to get rid of it entirely?

Dose units will be used as follows: ` enum DoseUnitsType { Unknown = -1, Gy = 0, Relative = 1 }; `

cpinter:
> Not quite sure what is the problem with levels table? Do you want to get rid of it entirely?

No, obviously not. But from your screenshot it seems that you have a new collapsible button within the Input section. The problem with that button is that it's not obvious where that section ends, and that it is unusual to have a collapsible button within another collapsible button. I suggest reorganizing the UI so that it is less confusing. Maybe instead of a collapsible button, you could put the "Prescribed or reference dose value" in a groupbox, but other solutions could work too.

Isodose/Logic/vtkMRMLIsodoseNode.h
MichaelColonel: There is no default value anymore.

Isodose/qSlicerIsodoseModuleWidget.cxx
MichaelColonel: Fixed.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
MichaelColonel: Two isolevels at the same time don't work quite right in hierarchy node. If both absolute and relative isolevels with the same postfix in hierarchy you can hide only one isolevel folder node, other can't be hidden. So if use only one postfix => only one isolevels folder node.

MichaelColonel:
> Maybe instead of a collapsible button, you could put the "Prescribed or reference dose value" in a groupbox, but other solutions could work too.

I've changed UI a little with some other fixes.

image

cpinter:
I like this much better, thank you!

@lassoan @gregsharp do you have anything to add? If not then I'll merge this on or after the weekend.

Issue #175, opened 2021-01-25
Installing SlicerRT makes Qt Designer crash


lassoan:
After SlicerRT is installed, Qt designer cannot be started, because it crashes in qSlicerBeamsModuleWidgetsPlugins.dll.

Qt Designer instantiates custom widgets (via Qt widget plugins). Since in this case the application is not derived from qSlicerCoreApplication, all calls to qSlicerCoreApplication::application() return nullptr. This is a likely root cause of the crash. You need to add null-pointer checks to all qSlicerCoreApplication::application() calls (at least to those that are called during widget initialization).

See how we have just fixed a similar issue in Slicer core:
https://github.com/Slicer/Slicer/commit/30f347f9f1da97be48222cfbf12c91d887278fc7

cpinter:
Designer now starts with SlicerRT. Thanks @MichaelColonel for the fix!

PR #176, opened 2021-01-26
BUG: Fix Qt designer crash because of BeamsModuleWidgetsPlugins

MichaelColonel:
MLC position logic doesn't check the qSlicerCoreApplication pointer during the initiation process into qMRMLBeamParametersTabWidgetPrivate.

See issue #175.


lassoan (commented):
Thank you, this looks good to me. Just to be consistent, it would be nice if you checked before all this->MLCPositionLogic access if it is nullptr or not (now in some methods it is checked, but in some other method it is not).

MichaelColonel:
Validation checks to the this->MLCPositionLogic pointer have been added.

PR #180, opened 2021-02-11
ENH: Combo boxed instead of radio buttons in plastimatch DRR parameters GUI, and threshold parameter

MichaelColonel:
HU threshold parameter has been added to process voxels with values higher than threshold. That was done for better bone visualization.


cpinter:
Looks good, thank you!

PR #182, opened 2021-03-24
Minor enhancements for future use

MichaelColonel:
- Added calculation of transformation for one frame to another for IEC standard. For example transformation from Patient Support frame to Collimator frame: Patient Support -> Fixed Reference -> Gantry -> Collimator. That will be used in RoomsEyeView module for static RAS data and dynamic treatment models.
- Added busy cursor for DRR calculation process.

### DIFF
--- a/Beams/Logic/vtkSlicerIECTransformLogic.cxx
+++ b/Beams/Logic/vtkSlicerIECTransformLogic.cxx
@@ -401,7 +401,8 @@ vtkMRMLLinearTransformNode* vtkSlicerIECTransformLogic::GetTransformNodeBetween(
 }
 
 //-----------------------------------------------------------------------------
-bool vtkSlicerIECTransformLogic::GetTransformBetween(CoordinateSystemIdentifier fromFrame, CoordinateSystemIdentifier toFrame, vtkGeneralTransform* outputTransform)
+bool vtkSlicerIECTransformLogic::GetTransformBetween(CoordinateSystemIdentifier fromFrame, CoordinateSystemIdentifier toFrame, 
+  vtkGeneralTransform* outputTransform, bool transformForBeam/* = true*/)
 {
   if (!outputTransform)
   {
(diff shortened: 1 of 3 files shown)


Beams/Logic/vtkSlicerIECTransformLogic.cxx
```diff
@@ -401,7 +401,8 @@ vtkMRMLLinearTransformNode* vtkSlicerIECTransformLogic::GetTransformNodeBetween(
 }
 
 //-----------------------------------------------------------------------------
-bool vtkSlicerIECTransformLogic::GetTransformBetween(CoordinateSystemIdentifier fromFrame, CoordinateSystemIdentifier toFrame, vtkGeneralTransform* outputTransform)
+bool vtkSlicerIECTransformLogic::GetTransformBetween(CoordinateSystemIdentifier fromFrame, CoordinateSystemIdentifier toFrame, 
+  vtkGeneralTransform* outputTransform, bool transformForBeam)
```
cpinter: We usually indicate the default value here for readability purposes. Can you change add this?
`bool transformForBeam/*=true*/`

DrrImageComputation/qSlicerDrrImageComputationModuleWidget.cxx
```diff
@@ -572,9 +572,14 @@ void qSlicerDrrImageComputationModuleWidget::onComputeDrrClicked()
     return;
   }
   
+  QApplication::setOverrideCursor(Qt::WaitCursor);
+
   bool result = d->logic()->ComputePlastimatchDRR( parameterNode, ctVolumeNode);
   if (result)
   {
+    // TODO: Restore previously selected node
```
cpinter: It seems to me that the TODO was resolved so it can be removed.

cpinter:
Thanks, looks good. I had only two minor comments.

MichaelColonel:
I've changed these minor comments.

PR #188, opened 2021-05-14
STYLE: Use cross platform vtkMath methods instead of 180./M_PI constant

MichaelColonel:
This should work in both Linux and Windows systems.


cpinter:
Please change the commit prefix from STYLE to ENH and we're good. Thank you!

MichaelColonel:
Prefix was changed, documentation updated.

New screenshots for module documentation:

image

image

cpinter:
Excellent, thank you once more!

PR #189, opened 2021-05-21
ENH: Added python test for DRR and documentation examples of how to compute DRR via python

MichaelColonel:


cpinter (approved):
Thank you, looks great!

Issue #80, opened 2018-01-21
Use real-time GPU isosurface rendering for in Isodose module


cpinter:
In addition to generating model nodes for the isosurfaces (which results in model hierarchies that are hard to keep in sync and hard to handle when multiple dose volumes are present), add support for GPU-based isosurface rendering. See Kitware blog post https://blog.kitware.com/gpu-rendering-of-isosurfaces/

* This should be the default way of showing isodose in 3D
* Explore whether this can be used for 2D
* Current method needs to be kept, in order to allow exporting of isodose surfaces

Related tickets:
https://github.com/SlicerRt/SlicerRT/issues/18
https://github.com/SlicerRt/SlicerRT/issues/8

MichaelColonel:
I'm going to add support for GPU isosurface volume rendering for 3D view as a part of VTK practice. According to [VTK example](https://kitware.github.io/vtk-examples/site/Cxx/VolumeRendering/RayCastIsosurface/) it shouldn't be very hard. 

A couple of questions beforehand:

1. Should i register a new mapper for volume rendering as it described in [docs](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#register-a-new-volume-rendering-mapper)?
2. Should i create a derived display node from [vtkMRMLGPUVolumeRayCastRenderingDisplayNode](http://apidocs.slicer.org/master/classvtkMRMLGPURayCastVolumeRenderingDisplayNode.html) or i can use that display node class directly?

lassoan:
As an experiment, you can just put the volume rendering actor and mapper directly in the VTK renderer (see helpful code snippets [here](https://slicer.readthedocs.io/en/latest/developer_guide/script_repository.html#access-vtk-views-renderers-and-cameras)).

The most important thing to try if it works well with multi-volume rendering (if you use a separate volume raycast mapper for each volume then they are not composited correctly). It would be completely useless if we could only render a single volume at a time.

In addition to this, you also need to test if point picking works on these surfaces (can you place a markup fiducial on the rendered surface?).

Even if everything works well, you would still need a solution for slice display. You would also need to implement non-linear transform support for volume rendering. Volume rendered surfaces cannot benefit from advanced physical based rendering options, screen-space ambient occlusion, surface textures and shaders.

[...]

PR #191, opened 2021-07-03
ENH: Represent calculated isodose surfaces as a segmentation node

MichaelColonel:
Toggle button near "Number of iso levels:" QSpinBox is added to GUI to select type of representation.

Original representation as models into folder
image

New representation as a single segmentation node
image


Isodose/Logic/vtkMRMLIsodoseNode.h
```diff
@@ -141,6 +146,9 @@ class VTK_SLICER_ISODOSE_LOGIC_EXPORT vtkMRMLIsodoseNode : public vtkMRMLNode
   /// Whether use relative isolevels representation
   /// for absolute dose (Gy) and unknown units or not
   bool RelativeRepresentationFlag;
+
+  /// Use segmentation flag for isodose models
+  bool SegmentationFlag;
```
cpinter: Quite unfortunate name I think, because only vaguely refers to its purpose. How about `UseSegmentationToStoreIsodoseSurfaces` ?

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
```diff
@@ -683,9 +693,49 @@ void vtkSlicerIsodoseModuleLogic::CreateIsodoseSurfaces(vtkMRMLIsodoseNode* para
...
+  }
+  else // Segmentation node
+  {
+    segmentationNode = vtkSmartPointer<vtkMRMLSegmentationNode>::New();
+    std::string segmentationNodeName = scene->GenerateUniqueName(isodoseFolderName.c_str());
+    segmentationNode->SetName(segmentationNodeName.c_str());
+    scene->AddNode(segmentationNode);
+
+    // Set master representation to planar contour
+    segmentationNode->GetSegmentation()->SetMasterRepresentationName(vtkSegmentationConverter::GetSegmentationPlanarContourRepresentationName());
```
cpinter: Can you please explain in detail how you intend to use these segmentation objects, including why you use planar contour as master?

Isodose/qSlicerIsodoseModuleWidget.cxx
```diff
@@ -922,6 +924,29 @@ void qSlicerIsodoseModuleWidget::updateButtonsState()
...
+  }
+
+  if (!toggled)
+  {
+    d->toggleButton_IsodoseRepresentation->setIcon(QIcon(":/Icons/Isodose.png"));
+    d->toggleButton_IsodoseRepresentation->setToolTip(tr("Represent isodoses as multiple model nodes into folder"));
+  }
+  else
+  {
+    d->toggleButton_IsodoseRepresentation->setIcon(QIcon(":/Icons/Segmentation.png"));
```
cpinter: Why do we have a button for this? I need to know more about this whole work, please give us an overview, and rather write more than less. Thanks.

MichaelColonel: It's a toggle button, if the button is pressed, then segmentation node would be used to store isodose surfaces. If the button isn't pressed, then original isodose representation would be used (model nodes into folder).

Toggle button is pressed in the red square.
image


Isodose surfaces are stored as a segmentation node named "RTDOSE[1]_IsodoseSurfaces"
image

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
MichaelColonel: It's an alternative option how to store isosurfaces.

the code:
` segmentationNode->SetMasterRepresentationToClosedSurface(); `
doesn't work. It doesn't show isodose surfaces on the views. It works with the planar contours.

Isodose/Logic/vtkMRMLIsodoseNode.h
MichaelColonel: I will rename it.

Isodose/qSlicerIsodoseModuleWidget.cxx
cpinter: This I understood of course. Why do you need to be able to switch between the two? Why expose it to the user? What is the reason you needed the segmentation representation in the first place?

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
lassoan: "Planar contours" are not suitable as master representation because they only store 2D contours and only in one specific orientation. Probably the best representation for storing isodose contours/surfaces is "closed surface".

Isodose/qSlicerIsodoseModuleWidget.cxx
lassoan: I think it would be enough to only use segmentation. It takes two clicks in Data module to export model hierarchy from a segmentation node.

This work is the result of this discussion: https://github.com/SlicerRt/SlicerRT/issues/80. @MichaelColonel add #80 to the commit comment to make it clear what issue this development resolves. I like this development, because it reduces the complexity of the scene (less nodes) and the visualization is nicer (filled color region and not just the contour).

It might be useful to have two modes:
- A. each segment contains a solid region of dose > threshold
- B. each segment contains  "ring" shaped region of thresholdA < dose < thresholdB

cpinter: I completely agree with these. I was mostly curious why do does this branch introduce this complexity of being able to switch from the module. Switching to segmentation representation instead of model nodes makes sense, but there was nothing in this PR referring to its motivation.

So I'd suggest removing the button and the logic for the switching. It adds unnecessary complexity as much on the UI as in the source code (making it more error-prone). Unless you have a good reason offering this option. Thanks!

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
cpinter: You can set the preferred display representations in the display node as far as I remember. That should fix the "doesn't show" part.

Isodose/qSlicerIsodoseModuleWidget.cxx
MichaelColonel: I will do as you suggested. At the beginning nobody told me to delete initial version of isodose representation, that is why i made a toggle button.

MichaelColonel: I added better description into commit, and implementation of two modes: Single border mode (A) and Double border mode (B). Mode can be switched before calculation by the toggle button.

PR #192, opened 2021-07-09
ENH: Changed the dsurround(1) from 10 cm to 60 cm to enable dose calculations for a source away from the phantom.

ThomasBulenga:
* We currently extend our research to include Co-60 beams for which the source is away from the phantom. The dsurround(1) of 10 cm works for Orthovoltage but limits dose calculation when dsource is increased as for the case of Co-60 source. Increasing the dsurround(1) accomodates Co-60 beams without affecting Orthovoltage dose calculations.

   
image


Sunderlandkyl:
Thanks @ThomasBulenga, could you change the commit message to use the prefix "ENH:"?

ThomasBulenga:
Done @Sunderlandkyl

Sunderlandkyl:
I think you just changed the title of the PR. It's Ok though, I changed the commit message when merging.

PR #194, opened 2021-07-30
ENH: Removed redundant modified signal event since built-in set methods already have modified signal

MichaelColonel:


cpinter:
Perfect, thank you!

PR #196, opened 2021-09-02
COMP: Collision detection compilation fix because of API changes

MichaelColonel:
SetInput method becomes [SetInputData](https://vtk.org/doc/nightly/html/classvtkCollisionDetectionFilter.html#aaabafe1a1ecae517c9559b2a88e1f519)


lassoan (approved):
Thank you! (I've updated these files but it seems that somehow I did not include them in my commit)

PR #200, opened 2021-10-22
Allow for setting fiducial specific RBF radius

simonoxen:
Hi, in the context of [this](https://discourse.slicer.org/t/plastimatch-landwarp-with-variable-rbf-radius/20226) discussion, would it be possible to include these changes?

It modifies landwarp allowing to specify multiple RBF radius. Default behavior is still one radius value.

This also requires the following changes in plastimatch source: SlicerRt/plastimatch#1

Thanks!

### DIFF
--- a/PlmLandwarp/plastimatch_slicer_landwarp.cxx
+++ b/PlmLandwarp/plastimatch_slicer_landwarp.cxx
@@ -99,7 +99,6 @@ main (int argc, char *argv[])
 
   /* Set other parameters */
   lw->default_val=plmslc_landwarp_default_value;
-  lw->rbf_radius=plmslc_landwarp_rbf_radius;
   lw->young_modulus=plmslc_landwarp_stiffness;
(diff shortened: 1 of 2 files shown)


lassoan (approved):
Thank you, this looks good to me.

simonoxen:
Hi, with [SlicerRt/plastimatch#1 ](https://github.com/SlicerRt/plastimatch/pull/1) not merged I ended up doing a CLI module that runs plastimatch-like algorithm for variable rbf radius ([https://github.com/netstim/SlicerNetstim/blob/master/fiducialRegistrationVariableRBF/fiducialRegistrationVariableRBF.cxx](https://github.com/netstim/SlicerNetstim/blob/master/fiducialRegistrationVariableRBF/fiducialRegistrationVariableRBF.cxx)).

Feel free to revert this commit, since it won't have any effects on the plastimatch call.

Thanks!

Issue #201, opened 2021-11-09
Problem with loading RTPLAN in Slicer version  4.13.0-2021-11-05 r30381


shahrokh2:
Hello Dear Developers and Users
As mentioned in Support of 3D Slicer Community, entitled [Problem with SlicerRT in Slicer version 4.11.20210226 r29738 / 7a593c8](https://discourse.slicer.org/t/problem-with-slicerrt-in-slicer-version-4-11-20210226-r29738-7a593c8/20510) (Nov 7), I have problem to load RTPLAN only in Slicer. 

## Steps to reproduce
1- I load CT images (**CT**), structures (**RTSRUCT**) and dose (**RTDOSE**) exported from **ISOgray®**  (a Treatment Planning System for external radiotherapy) completely and without any problems.
2- I can not load beams (**RTPLAN**) to Slicer. When doing it in _DICOM_ module, I get this message in the window entitled **DICOM loading”: Could not load: 4: RTPLAN: Beamsetup1CTDB as a RT**.
3- Also, I ticked _Advanced_, in _DICOM database_ and then click on _Examine_ key. I get this message in the table of _Warnings_: **Reference image in series does not contain geometry information. Please use caution…**
4- Anonymized data that I have trouble loading (including as RTDOSE, RTSTRUCT, CT (74 images) and RTPLAN) can be downloaded via the link below:
[https://www.dropbox.com/sh/e1d0yc24ghptslk/AADZyW550Q6geYbDuRn3hAMna?dl=0](url)

At this anonymized data, as mentioned above I can load  RTDOSE, RTSTRUCT, CT and **can not load  RTPLAN**.

I have exactly same problem with other versions of 3DSlicer except **4.10.2 r28257**. In other words, I must mentioned that I can load CT images, RTPLAN, RTSTRUCT and RTDOSE completely and without any problems in Slicer version: 4.10.2 r28257.

## Expected behavior
I want to work with new options added to SlicerRT module in last version of Slicer (version 4.13.0-2021-11-05 r30381 / f46c474) after loading data exported from our TPS (ISOgray).

## Environment
OS:                 Ubuntu 18.04.4 LTS
OS Type:	64-bit
Processor: 	Intel® Core™ i7-9700K CPU @ 3.60GHz × 8 
Graphics:	GeForce RTX 2070 SUPER/PCIe/SSE2
Slicer:		4.13.0-2021-11-05 r30381 / f46c474
SlicerRT	         Installed from Extension Manager

[...]

PR #202, opened 2021-11-09
ENH: Better check of beams module and fraction scheme module presence

MichaelColonel:
Has been added a better check of beams module and fraction scheme module presence in RTPLan. Fix for issue #201 

The [standard](http://dicom.nema.org/medical/dicom/current/output/chtml/part03/sect_A.20.3.html) describe several cases of relation between beams and fraction scheme module:
1. Only beams module  (brachy application setups must be absent)
2. Both beams and fraction scheme modules, number of beams is greater than zero for one or more fraction groups (brachy application setups must be absent)
3. Only brachy application setups module (beams module must be absent)
4. Both brachy and fraction scheme modules, number of brachy application setups is greater than zero for one or more fraction groups (beams module must be absent)

### DIFF
--- a/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
+++ b/DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
@@ -802,6 +802,7 @@ void vtkSlicerDicomRtReader::vtkInternal::LoadRTPlan(DcmDataset* dataset)
   }
 
   vtkDebugWithObjectMacro(this->External, "LoadRTPlan: Load RT Plan object");
+  bool hasFractionSchemeModule = true;
   if (rtPlan.isRTFractionSchemeModulePresent(OFTrue) == OFTrue)
   {
     vtkDebugWithObjectMacro( this->External, "LoadRTPlan: Fraction Scheme is correct");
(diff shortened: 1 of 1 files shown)


DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -816,8 +817,8 @@ void vtkSlicerDicomRtReader::vtkInternal::LoadRTPlan(DcmDataset* dataset)
   }
   else
   {
-    vtkErrorWithObjectMacro( this->External, "LoadRTPlan: Fraction Scheme is absent");
-    return;
+    vtkWarningWithObjectMacro( this->External, "LoadRTPlan: Fraction Scheme is absent");
```
lassoan: Is this an error that fraction scheme is absent?
If not then it is better to log it with a debug macro.

lassoan (commented):
Thanks a lot for the quick fix. I just added a question in the code.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: Fixed

lassoan (approved):
Thank you for the quick fix, this looks good now.

cpinter:
Thanks for the quick fix Mikhail!

PR #205, opened 2022-01-28
BUG: Fix crash while loading RTDOSE DICOM file

MichaelColonel:
1. Fix crash while loading RTDOSE DICOM file;
2. Fix deprecated Markups API warning;
3. Minor changes.

"Isodose" module logic creates several color tables. In order to change or add data into color table, for isodose and dose visualization, it must be editable or a copy of a default color table. If the scene was cleared and all the nodes were deleted, the editable isodose was deleted as well, and it wouldn't be created again with proper lookup table.

For that reason if one is trying to load RTDOSE volume, the program crashes at the [line](https://github.com/SlicerRt/SlicerRT/blob/master/DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx#L571-#L572), even if isodose table node pointer is valid!

А singleton tag was added to the copy of default isodose color table to fix the crash.

How to test the crash.

1. Load a DICOM series with RTDOSE, for example: [https://github.com/SlicerRt/SlicerRtData/tree/master/eclipse-8.1.20-phantom-prostate](https://github.com/SlicerRt/SlicerRtData/tree/master/eclipse-8.1.20-phantom-prostate)
2. Clear the scene. "File"->"Close Scene" or Ctrl+W -> "Close Scene"
3. Try to load the same series once again.

### DIFF
--- a/Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
+++ b/Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
@@ -111,23 +111,20 @@ void vtkSlicerIsodoseModuleLogic::SetMRMLSceneInternal(vtkMRMLScene* newScene)
   // Create a copy of default isodose color table. The copy can be edited.
   if (isodoseColorTableNode)
   {
-    // Get "Colors" module logic
-    auto aLogic = this->GetApplicationLogic()->GetModuleLogic("Colors");
-    vtkSlicerColorLogic* colorLogic = vtkSlicerColorLogic::SafeDownCast(aLogic);
-
     // Create a copy of isodose color table with unique name
-    std::string& uniqueName = vtkSlicerIsodoseModuleLogic::IsodoseColorNodeCopyUniqueName;
-    uniqueName = this->GetMRMLScene()->GenerateUniqueName(isodoseColorTableNode->GetName());
-    vtkMRMLColorTableNode* colorNode = colorLogic->CopyNode(isodoseColorTableNode, uniqueName.c_str());
+    std::string uniqueName = this->GetMRMLScene()->GenerateUniqueName(vtkSlicerIsodoseModuleLogic::IsodoseColorNodeCopyUniqueName.c_str());
+    // Get "Colors" module logic and copy a isodose color node
+    vtkMRMLColorTableNode* colorNode = this->GetMRMLApplicationLogic()->GetColorLogic()->CopyNode(isodoseColorTableNode, uniqueName.c_str());
     if (colorNode)
     {
+      colorNode->SetSingletonTag(DEFAULT_ISODOSE_COLOR_TABLECOPY_NODE_NAME); // set tag that it's a copy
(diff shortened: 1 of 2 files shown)


Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
```diff
@@ -296,15 +293,24 @@ vtkMRMLColorTableNode* vtkSlicerIsodoseModuleLogic::GetDefaultIsodoseColorTable(
       vtkWarningWithObjectMacro(scene, "GetDefaultIsodoseColorTable: Multiple default isodose color table nodes found");
     }
 
-    vtkMRMLColorTableNode* isodoseColorTableNode = vtkMRMLColorTableNode::SafeDownCast(defaultIsodoseColorTableNodes->GetItemAsObject(0));
-    return isodoseColorTableNode;
+    vtkObject* object = defaultIsodoseColorTableNodes->GetItemAsObject(0);
+    if (object)
```
cpinter: I think if number of items is greater than 0 then the 0th object will always be a valid vtkObject. The only thing we need to check is its type. However, if the type is not the one you want to SafeDownCast to, then the `isodoseColorTableNode` pointer will be null, so it will return null. I personally would remove this particular check in order to keep the file shorter and thus more readable.

cpinter:
Thanks @MichaelColonel ! It's not obvious to me which change exactly fixes the crash. Also not sure about the difference between DefaultCopy and DefaultUserCopy. Can you be a bit more verbose in the commit message please?

cpinter:
Also please confirm there are no new tests failing.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
MichaelColonel: I've restored origin part.

cpinter:
Are you still working on it, or ready to review?

MichaelColonel:
I've updated a commit commentary and it's ready for another review. I haven't run tests yet.

MichaelColonel:
`ctest -C Release`

gives
```
The following tests FAILED:
	  1 - qSlicerBeamsModuleGenericTest (Failed)
	  2 - qSlicerBeamsModuleWidgetGenericTest (Failed)
	  4 - qSlicerIsodoseModuleGenericTest (Failed)
	  5 - qSlicerIsodoseModuleWidgetGenericTest (Failed)
	  6 - vtkSlicerIsodoseModuleLogicTest_EclipseProstate (Failed)
	 12 - py_DicomRtImportTest (Failed)
	 47 - py_nomainwindow_PlmProtonDoseEngineTest (Failed)
	 54 - py_BatchStructureSetConversion (Failed)
	 62 - py_DrrImageComputationTest (Failed)
```

MichaelColonel:
Python scripts are failing because of data URL is no longer valid, and test data is unavailable.

cpinter:
Thank you so much, it looks all good to me now.

PR #210, opened 2022-03-28
BUG: Fix failing vtkSlicerIsodoseModuleLogicTest_EclipseProstate

Sunderlandkyl:
The isodose surface generated with VTK9 is slightly different around the holes in the bottom of the mesh. This causes a small volume difference between the baseline and generated surface. The surface is the same in all other areas, so there doesn't seem to be an issue. Fixed by updating the baseline.

Shown below, green areas are where the VTK9 mesh deviates from the baseline:
image


Sunderlandkyl:
@lassoan @cpinter Just checking to make sure that everyone is ok with updating the test data.

Issue #208, opened 2022-03-21
Fix Isodose module crash by using color legend instead of scalar bars


MichaelColonel:
Isodose module crashes if any "Show scalar bar" check box is toggled.
Scalar bars that show isodose levels in the "Isodose" module must be replaced with color legend.

cpinter:
Seems to be fixed, closing.

PR #211, opened 2022-03-31
ENH: Use color legend instead of scalar bars

MichaelColonel:
This is a possible solution for #208.
1. Dummy vtkMRMLDisplayableNode and vtkMRMLDisplayNode were added to vtkMRMLIsodoseNode to allow visualization of color legend with vtkMRMLColorLegendDisplayNode;
2. Scalar bars are replaced with color legend;
3. qMRMLColorLegendDisplayNodeWidget was added to control color legend parameters.

image


lassoan:
Adding dummy nodes is not very clean and makes it more difficult to synchronize the visibility  of the isodose surfaces and the color legend. 

Instead, a nicer solution would be to use the subject hierarchy folder's existing vtkMRMLFolderDisplayNode for this.

We could also consider creating a single model node for all the isodose surfaces and color them using point scalars. It would simplify many other things, too. The only thing that would be harder is to quickly toggle visibility of individual lines/surfaces in the data module, but it might not be that important (the levels can be edited in isodose module).

@cpinter what do you think about changing the design to keep all the isodose lines in a single model?

cpinter:
I think it would lead to simpler code in most places.

Yes the only thing indeed is individual handling of the levels. I'm not sure how often it is needed by actual users, but if @MichaelColonel and @gregsharp say that almost never, then it's probably clear and we don't have to worry about it. Otherwise we could add a feature that separates the surfaces by the scalars.

MichaelColonel:
I have no objections, but what do you mean by coloring with point scalars? Something like: https://kitware.github.io/vtk-examples/site/Cxx/PolyData/ColoredPoints/, or am i wrong?

cpinter:
Yes, this is the mechanism. You have an array that assigns some value to each point (or cell - that is also possible) in a way that the point indices match the value indices in the array, then you set the array with `GetPointData()->SetScalars`. In the models module you can color the polydata using scalars by wnabling it in the Display/Scalars section.

MichaelColonel:
I've redone it with a single vtkMRMLModelNode.

Isodose/CMakeLists.txt
```diff
@@ -42,9 +46,12 @@ set(MODULE_UI_SRCS
 )
 
 set(MODULE_TARGET_LIBRARIES
+  vtkSlicer${MODULE_NAME}ModuleMRML
   vtkSlicer${MODULE_NAME}ModuleLogic
-  vtkSlicer${MODULE_NAME}ModuleWidgets
+#  vtkSlicer${MODULE_NAME}ModuleWidgets
```
cpinter: Why was it needed to comment this? The include directlry is still added, so removing the lib is inconsistent.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
```diff
@@ -62,14 +62,19 @@
...
+
 #include "vtksys/SystemTools.hxx"
 
 //----------------------------------------------------------------------------
 const char* DEFAULT_ISODOSE_COLOR_TABLE_FILE_NAME = "Isodose_ColorTable.ctbl";
 const char* DEFAULT_ISODOSE_COLOR_TABLE_NODE_NAME = "Isodose_ColorTable_Default";
 const char* DEFAULT_ISODOSE_COLOR_TABLECOPY_NODE_NAME = "Isodose_ColorTable_DefaultCopy";
 const char* RELATIVE_ISODOSE_COLOR_TABLE_NODE_NAME = "Isodose_ColorTable_Relative";
-const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_MODEL_NODE_NAME_PREFIX = "IsodoseLevel_";
+const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_MODEL_NODE_NAME_PREFIX = "IsodoseLevels";
```
cpinter: This is not used as a prefix any more although the name suggests so. I think it would be good to use it as a name postfix, i.e. you append "_IsodoseLevels" to the dose volume name so that they can be differentiated when there is more than one dose (the current way all of them have the exact same name). Make sure you overwrite any existing one.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
```diff
@@ -62,14 +62,19 @@
...
 const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_PARAMETER_SET_BASE_NAME_PREFIX = "IsodoseParameterSet_";
 const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_ROOT_HIERARCHY_NAME_POSTFIX = "_IsodoseSurfaces";
 const std::string vtkSlicerIsodoseModuleLogic::ISODOSE_RELATIVE_ROOT_HIERARCHY_NAME_POSTFIX = "_RelativeIsodoseSurfaces";
```
cpinter: It seems this one is not used anymore. Please remove all unused pieces of code.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
```diff
@@ -880,10 +799,57 @@ void vtkSlicerIsodoseModuleLogic::CreateIsodoseSurfaces(vtkMRMLIsodoseNode* para
...
+    displayNode->SetActiveScalarName("isolevels");
+    displayNode->SetAutoScalarRange(true);
+    displayNode->SetAndObserveColorNodeID(colorTableNode->GetID());
+    displayNode->SetScalarVisibility(true);
+
+    // Disable backface culling to make the back side of the model visible as well
+    displayNode->SetBackfaceCulling(0);
+
+    vtkNew<vtkMRMLModelNode> isodoseModelNode;
+    std::string uniqueName = scene->GenerateUniqueName(vtkSlicerIsodoseModuleLogic::ISODOSE_MODEL_NODE_NAME_PREFIX);
```
cpinter: This is where you create the ambiguous name. Please use the dose volume's name as prefix and just add a postfix saying this is the isodose levels model.

cpinter (changes requested):
Thanks for the contribution! In general it looks great. I added a few minor comments.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
MichaelColonel: > Make sure you overwrite any existing one.

So it must be only one isodose model node per dose volume?

cpinter: That was how it has worked so far. It is very easy to generate an isodose set, and it is not useful to have them parallel I think, so yes.
@lassoan @Sunderlandkyl what do you think?

lassoan: When there were many output isodose models then things were already complicated so it made some sense to not make things even more complicated by having multiple sets of isodose models per dose volume.

Now that we will have a single output model, we can follow the usual output node selection pattern: have a node selector for the output model and overwrite that node. If the user wants to create/update another output then he can create/choose a different output model. This would make the Isodose module behave like most other modules.

MichaelColonel: Should the isodose model selector be separated from input parameters? For example Input and Output collapsible buttons.

image

lassoan: Yes, I think it looks good. Maybe we could make the "Generated isodose" button a but larger (wider, higher). This button looks like a QToolButton (which is much more tight around the text) and changing it to a QPushButton could make it a bit larger, more prominent.

The output isodose model could have a "None" option with custom None text "Create new isodose model", so if the user leaves it at default then a new model is created (after the model is created it would be set as output model).

Tiny remark: we generally try to avoid using "node" on the user interface, could you replace "Isodose model node" to "Model node"?

lassoan: It would be nice to add the unit to the legend. I would also make things look a bit more pleasing if the size of 5, 10, 15, ... numbers would be a bit smaller and the title of the colorbar would be displayed closer to the bar.

MichaelColonel: > I would also make things look a bit more pleasing if the size of 5, 10, 15, ... numbers would be a bit smaller and the title of the colorbar would be displayed closer to the bar.

That is a bug in color legend actor when label == text.

lassoan: Yes, indeed there seems to be some unnecessary cross-talk between width of a vertical color legend and the height of the colorbar: if the width is increased then the height of the colorbar is becomes wider and shorter. Overall, the behavior seems to be about the same, regardless of what is used as label (value or color name). Not a big issue, not urgent to fix it, but at some point it would be nice to investigate this.

MichaelColonel: I've added a method to setup color legend defaults, when color legend was created for isodose model for the first time.

image

lassoan: This looks nice. If you load images from DICOM and slice view annotations are displayed then the top-left and bottom-left corners can contain many lines of patient information. There may be some space on the left side near the middle, but if we want to avoid overlapping of patient information and color legend then it would be safer to display color legend on the right side by default.

MichaelColonel: Position was changed to the right side of the screen.

MichaelColonel: Color legend in the right side of the screen by default.

image

MichaelColonel:
I have updated an old PR, so it can be merged with current master. Everything should work out of the box.

Isodose/Logic/vtkSlicerIsodoseModuleLogic.h
```diff
@@ -60,20 +57,20 @@ class VTK_SLICER_ISODOSE_LOGIC_EXPORT vtkSlicerIsodoseModuleLogic : public vtkSl
...
 
-  /// Accumulates dose volumes with the given IDs and corresponding weights
-  void CreateIsodoseSurfaces(vtkMRMLIsodoseNode* parameterNode);
-
-  /// Get isodose folder for a dose volume
-  /// \param node Dose volume node or isodose parameter node referencing the dose volume
-  /// \return Subject hierarchy item ID of the folder containing the isodose surfaces. 0 if not found
-  vtkIdType GetIsodoseFolderItemID(vtkMRMLNode* node);
+  /// Create dose isolevels surfaces for dose volume associated with the parameterNode
+  /// @param parameterNode - isodose node parameters
```
cpinter: Please use the same notion as everywhere else: `\param` and no dash

cpinter (changes requested):
This is a much needed modernization. Thanks a lot!

I only had time to quickly read the changes (and had only one very minor request), and could not test either. If you say it works perfectly I believe you and can be merged after that tiny fix.

It would probably be useful if @Sunderlandkyl or @lassoan can review it too.

Isodose/MRML/vtkMRMLIsodoseNode.cxx
```diff
@@ -88,10 +88,8 @@ void vtkMRMLIsodoseNode::ReadXMLAttributes(const char** atts)
   vtkMRMLReadXMLBeginMacro(atts);
   vtkMRMLReadXMLBooleanMacro(ShowIsodoseLines, ShowIsodoseLines);
   vtkMRMLReadXMLBooleanMacro(ShowIsodoseSurfaces, ShowIsodoseSurfaces);
-  vtkMRMLReadXMLBooleanMacro(ShowScalarBar, ShowScalarBar);
-  vtkMRMLReadXMLBooleanMacro(ShowScalarBar2D, ShowScalarBar2D);
   vtkMRMLReadXMLBooleanMacro(ShowDoseVolumesOnly, ShowDoseVolumesOnly);
-  vtkMRMLReadXMLIntMacro(DoseUnits, DoseUnits);
+  vtkMRMLReadXMLEnumMacro(DoseUnits, DoseUnits);
```
Sunderlandkyl, line 92: Wouldn't changing the Read/Write DoseUnits macro from Int to Enum break compatibility with existing saved scenes?

MichaelColonel: I will check that.

lassoan: Backward compatibility (new versions of SlicerRT can read both old and new scenes) can be maintained by implementing the `Get##propertyName##FromString(xmlReadAttValue)` method so that it recognizes old integer values.

This breaks future compatibility (scenes that are created in new version of SlicerRT will not be fully readable in older versions), but this may be acceptable, especially since extensions auto-update is now available in Slicer5 (making it easier for users to keep SlicerRT up-to-date).

MichaelColonel:
The method `int GetDoseUnitsFromString(const char *)` has already been added:
https://github.com/MichaelColonel/SlicerRT/blob/isodose-cl/Isodose/MRML/vtkMRMLIsodoseNode.h#L113
https://github.com/MichaelColonel/SlicerRT/blob/isodose-cl/Isodose/MRML/vtkMRMLIsodoseNode.cxx#L236-L256

So the scene with should be loaded, but if you recalculate isodose with color legend, the isolines will be too thin and barely visible.

lassoan:
> The method `int GetDoseUnitsFromString(const char *)` has already been added:

That's good, but for backward compatibility `GetDoseUnitsFromString` should recognize the old values (`0`, `1`, and `-1`) and not just the new strings (`Gy`, `Relative`, and `Unknown`).

> the isolines will be too thin and barely visible

Why the isolines thickness is related to the dose units?

MichaelColonel:
> That's good, but for backward compatibility `GetDoseUnitsFromString` should recognize the old values (`0`, `1`, and `-1`) and not just the new strings (`Gy`, `Relative`, and `Unknown`).

Method modified.

I check the process of loading scene, and the DoseUnits is preserved between an old isodose color bar and a new color legend.

Absolute Dose scene: https://disk.yandex.ru/d/iVwup0TxZoiGXw
Relative Dose scene: https://disk.yandex.ru/d/G9q9TB6Z6uDyVA

> Why the isolines thickness is related to the dose units?
Correct. It isn't related, but the difference in visualization is noticeable.

MichaelColonel:
Other requested changes will be corrected shortly.

MichaelColonel:
Doxygen description has been [updated](https://github.com/MichaelColonel/SlicerRT/commit/05bc557d676af54b06c295b4cb883f2ec226f0f2).

cpinter:
Sorry I'm just back from vacation. Anyting to do before integrating this PR?

MichaelColonel:
I will check everything one more time and let you know.

MichaelColonel:
It's ready to be merged.

cpinter:
Thanks @MichaelColonel ! GitHub tells me the branch cannot be rebased due to conflicts, can you please manually rebase the branch?

@lassoan @Sunderlandkyl any requests before merging?

Sunderlandkyl:
No, lgtm!

lassoan:
OK to merge for me, too. I would merge it, but, as @cpinter noted, there are conflicts that @MichaelColonel would need to resolve by rebasing on latest master.

MichaelColonel:
@cpinter could you try to merge it? I hope there are no more conflicts.

lassoan:
We still have the error message: "This branch cannot be rebased due to conflicts"

- Pull from SlicerRt:master
- Rebase
- Force-push to MichaelColonel:isodose-cl

MichaelColonel:
I've rebased it.

lassoan:
It is good now, merged. Thanks a lot for your contribution!

Issue #216, opened 2022-09-14
Update ExternalBeamPlanning module to use markup ROI instead of annotation ROI


lassoan:
Annotations module, which provides `vtkMRMLAnnotationROI` and `vtkMRMLAnnotationRuler` nodes have been deprecated since April 2021 and will be removed in Slicer-5.4 stable version early next year.

ExternalBeamPlanning module uses annotation ROI node in ExternalBeamPlanning\Resources\UI\qSlicerExternalBeamPlanningModule.ui.  This module need to be updated to use markups nodes before the end of this year.

[This section](https://www.slicer.org/wiki/Documentation/Nightly/Developers/Tutorials/MigrationGuide/Slicer#Slicer_5.3%3A_Removed_Annotation_module) of the migration guide should help with implementing the changes and any questions can be asked at the [Remove Annotations module topic in the Slicer forum](https://discourse.slicer.org/t/remove-annotations-module/25271).

ferdymercury:
@lassoan was this already fixed by https://github.com/SlicerRt/SlicerRT/pull/217 and can thus be closed?

PR #217, opened 2022-09-15
ENH: Use vtkMRMLMarkupsROINode instead of vtkMRMLAnnotationROINode in ExternalBeamPlanning module

MichaelColonel:
Fix for #216. The DoseROI isn't implemented in any case in ExternalBeamPlanning module.


lassoan (approved):
Thank you!

PR #221, opened 2022-11-12
ENH: Calculate perspective projection of a point on the DRR imager plane

MichaelColonel:
SlicerRT user requested some update for DRR Image Computation module. See https://discourse.slicer.org/t/optimizing-resolution-and-spacing-parameters-in-plastimatch-drr/25873


lassoan (approved):
Thank you, it looks good.

PR #222, opened 2022-11-23
ENH: GUI upgrade of perspective projection calculation:

MichaelColonel:
* QTableWidget to represent projected control points;
* Project only selected or all control points;
* Flag to created a new markups fiducial node for projected points.

### DIFF
--- a/DrrImageComputation/Logic/vtkSlicerDrrImageComputationLogic.cxx
+++ b/DrrImageComputation/Logic/vtkSlicerDrrImageComputationLogic.cxx
@@ -34,6 +34,7 @@
 #include <vtkMRMLMarkupsPlaneNode.h>
 #include <vtkMRMLMarkupsFiducialNode.h>
 #include <vtkMRMLMarkupsLineNode.h>
+#include <vtkMRMLTableNode.h>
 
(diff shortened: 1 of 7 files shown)


MichaelColonel:
New interface

image

cpinter:
After a very quick glance it looks OK to me. One question so far: was there a specific reason for removing `ModuleWindowInitialized`? Thanks!

MichaelColonel:
Module is quite simple and it doesn't require any specific GUI initiation. The GUI doesn't show beam node correctly if parameterNode was created in python console if you switch into module GUI for the first time with `ModuleWindowInitialized`.

cpinter:
Makes sense, thanks! I won't be able to do a thorough review, but since you are the main user/developer of the DRR module I'd simply accept this change. @Sunderlandkyl @lassoan please review if interested. I'll merge it in a day or two (unless you need it merged asap @MichaelColonel in which case I can do it now). Thanks again!

MichaelColonel:
I thick it could be merged, so it could be tested by the user.

lassoan:
Thank you for all your hard work.

PR #223, opened 2022-11-29
ENH: Add real-time mode to Isodose computation

cpinter:
Real-time mode can be enabled using the RealTime flag in the Isodose parameter node. When this flag is enabled, the following functions are prevented: use of subject hierarchy to organize the isodose model nodes, reporting of progress, batch processing, and update of dose color table from the isodose one. Instead, top-level isodose model nodes are re-used (by node name) at every computation. It is useful when isodose is needed to be computed on-the-fly for streamed dose data, when one set of isodose surfaces (the "current one") is all that is needed to be kept and displayed.

### DIFF
--- a/Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
+++ b/Isodose/Logic/vtkSlicerIsodoseModuleLogic.cxx
@@ -613,14 +613,21 @@ bool vtkSlicerIsodoseModuleLogic::CreateIsodoseSurfaces(vtkMRMLIsodoseNode* para
     return false;
   }
 
-  scene->StartState(vtkMRMLScene::BatchProcessState);
+  if (!parameterNode->GetRealTime())
+  {
+    scene->StartState(vtkMRMLScene::BatchProcessState);
+  }
 
   // Get subject hierarchy item for the dose volume
-  vtkIdType doseShItemID = shNode->GetItemByDataNode(doseVolumeNode);
-  if (!doseShItemID)
+  vtkIdType doseShItemID = 0
+  if (!parameterNode->GetRealTime())
   {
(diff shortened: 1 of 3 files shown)


cpinter:
Let me fix the conflicts. Unfortunately in the project I added this feature for the latest SlicerRT did not compile on Ubuntu with gcc due to commit https://github.com/SlicerRt/SlicerRT/commit/220f766a011a15444da3531885b4fec2b5944384 and because of time constraints I could not deal with that yet.

cpinter:
I resolved the conflicts but have some doubts about how much sense it makes now that there is only one isodose model node. The big time sink for real-time use (showing isodose while streaming) was the batch processing and the SH operations (deletions, reparentings). I'd like to test this branch with the real-time use case it is intended for.

@Sunderlandkyl do you have any ideas why the said commit (see in previous comment) causes build errors with gcc? It is OK on the dashboard and I don't have any clear idea, maybe you have a hunch. Otherwise I'll investigate once I get to it. Thanks!

cpinter:
By the way it may make sense to remove batch processing in any case, now that only one node is being touched.

Sunderlandkyl:
Not sure why https://github.com/SlicerRt/SlicerRT/commit/220f766a011a15444da3531885b4fec2b5944384 would cause a build failure on gcc, what is the error? The commit itself just follows the changes made by this commit in Slicer: https://github.com/Slicer/Slicer/commit/693c81929807517756fab1d69c65f24da787989e. Is the gcc build using up-to-date Slicer source code?

cpinter:
Closing PR as the feature had to be reimplemented due to the changes in upstream. Will reopen a new PR when it is ready.

Issue #224, opened 2022-12-07
Isodose calculation replaces previously created isodose


ywatana1:
With older versions, I could overlay isodose lines of two RTDOSE files using the isodose tool. But, with the latest version 5.3.0-2022-11-26, I cannot do it anymore. It replaces the isodose line plot when I generate isodose lines after changing the RTDOSE file.

MichaelColonel:
Dear @ywatana1,

Do you need one isodose lines model node per RTDOSE volume or multiple isodose model nodes?

ywatana1:
I am not sure what you mean by "isodose model nodes". What I need it is to overlay a set of isodose lines from RTDOSE #2 on the dose from RTDOSE#1, resulting in two sets of isodose lines on one display. I use this function to compare two dose calculation methods.

MichaelColonel:
Possible fix is ready. PR #225.

cpinter:
Integrated. @ywatana1 please test tomorrow (when the Slicer build system provides a new build that includes this change) and close if it works as expected. Thank you!

ywatana1:
Great. The functionality back on the 1/5/2023 version. 
If we could select different line types of each set, it would be even better. 
Thank you very much.

MichaelColonel:
Good, that it works as it should.

> If we could select different line types of each set, it would be even better.

I will think how that feature can be implemented.

PR #225, opened 2022-12-31
BUG: A newly created isodose lines node will not replace an old one

MichaelColonel:
Possible fix for issue #224. Multiple isolines model nodes for a single RTDOSE, only the last isolines model node is observed.


cpinter:
Thank you very much and sorry for the wait

PR #230, opened 2023-05-30
COMP: remove use of constants defined in vtkSlicerVersionConfigure.h

achataigner:


lassoan (approved):
Thanks for working on this, it looks good to me.

The old Slicer wiki will be shut down soon. When we move the documentation to github then this will need to be revisited.

PR #238, opened 2023-11-15
BUG: Fix wrong DRR orientation after recent IEC logic changes

MichaelColonel:
Small fix for RtImage orientation after recent IEC logic changes.


cpinter:
Thank you!

Do you have a concrete suspicion about the bug you're mentioning in the comment?

MichaelColonel:
I don't. I'm going to rework some parts of logic to add support of orientated volumes. Nowadays DRR modules only supports standard DICOM orientation (LPS). Possibly that plastimatch part will required some work as well.

cpinter:
Thanks for the quick answer! I don't have the means to test this change but will merge now, especially that you say that you'll keep working on this part.

Issue #250, opened 2024-07-16
Fix DRR module


cpinter:
After the recent IEC transform logic refactoring need to revise the DRR module and fix any problems.

See https://github.com/SlicerRt/SlicerRT/pull/249

PR #251, opened 2024-07-16
BUG: Fix DRR image vup and normal vectors update after recent IEC logic refactoring

MichaelColonel:
There are minor changes in logic file, where we are updating view-up and normal vectors of the imager.
Fix for PR #250.

### DIFF
--- a/Beams/Logic/vtkSlicerIECTransformLogic.cxx
+++ b/Beams/Logic/vtkSlicerIECTransformLogic.cxx
@@ -30,6 +30,8 @@
 
 // STD includes
 #include <array>
+#include <algorithm>
+
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


Beams/Logic/vtkSlicerIECTransformLogic.cxx
```diff
@@ -30,6 +30,8 @@
 
 // STD includes
 #include <array>
+#include <algorithm>
```
cpinter, line 33: Why this is needed? We didn't have any build error. Was it a problem on Linux?

cpinter:
Integrated after a minor fix in 7168e016ddaa6646ac97b0a9ef5f1cc5104a90a0

Thank you for the quick fix!

PR #236, opened 2023-10-21
ENH: Fix patient support rotation angle to fixed reference frame

MichaelColonel:
Fix for issue #235.

PR #255, opened 2024-10-10
BUG: Fix patient support rotation angle to fixed reference

MichaelColonel:
How to test:

1. Load [https://github.com/SlicerRt/SlicerRtData/tree/master/xio-4.64-phantom-prostate-beams-allnonzero](https://github.com/SlicerRt/SlicerRtData/tree/master/xio-4.64-phantom-prostate-beams-allnonzero) from SlicerRtData. That dataset has predefined nonzero angles of beam orientation: gantry, couch, collimator.

2. Check that dose distribution lays within beam's model borders.

3. In "Beams" module check that beam orientation parameters are as follows:
  Gantry angle = 20 deg;
  Collimator angle = 350 deg;
  Couch angle = 300 deg.

4. In "Room's Eye View" module, load Varian machine and set required parameters for the machine sliders:
  Gantry angle = 20 deg;
  Collimator angle = 350 deg;
  Patient support rotation = -60 deg. (since slider range is from -180 up to 180 deg)

5. Go back to the "Beams" module and check that beam orientation parameters are the same:
  Gantry angle = 20 deg;
  Collimator angle = 350 deg;
  Couch angle = 300 deg.

### DIFF
--- a/Beams/Logic/vtkSlicerBeamsModuleLogic.cxx
+++ b/Beams/Logic/vtkSlicerBeamsModuleLogic.cxx
@@ -397,7 +397,9 @@ void vtkSlicerBeamsModuleLogic::UpdateIECTransformsFromBeam(vtkMRMLRTBeamNode* b
   // Set beam angles to IEC logic
   this->IECLogic->UpdateGantryToFixedReferenceTransform(beamNode->GetGantryAngle());
   this->IECLogic->UpdateCollimatorToGantryTransform(beamNode->GetCollimatorAngle());
-  this->IECLogic->UpdatePatientSupportRotationToFixedReferenceTransform(beamNode->GetCouchAngle());
+  // Set the inverse of the couch angle as now what moves is the room (treatment machine) around the patient,
+  // so the patient support table top (couch) always stays stationary in RAS.
+  this->IECLogic->UpdatePatientSupportRotationToFixedReferenceTransform(-1. * beamNode->GetCouchAngle());
 
   // Update fixed reference to RAS transform as well
   vtkMRMLRTPlanNode* parentPlanNode = beamNode->GetParentPlanNode();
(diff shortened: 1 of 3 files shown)


MichaelColonel:
That bug was fixed in #236 commit, but appeared once again when IEC logic had been reworked. I don't know how that has happened, but it should work again.

Beams/Logic/vtkSlicerBeamsModuleLogic.cxx
```diff
@@ -397,7 +397,7 @@ void vtkSlicerBeamsModuleLogic::UpdateIECTransformsFromBeam(vtkMRMLRTBeamNode* b
   // Set beam angles to IEC logic
   this->IECLogic->UpdateGantryToFixedReferenceTransform(beamNode->GetGantryAngle());
   this->IECLogic->UpdateCollimatorToGantryTransform(beamNode->GetCollimatorAngle());
-  this->IECLogic->UpdatePatientSupportRotationToFixedReferenceTransform(beamNode->GetCouchAngle());
+  this->IECLogic->UpdatePatientSupportRotationToFixedReferenceTransform(-1. * beamNode->GetCouchAngle());
```
ferdymercury, line 402: If signs are flipped here, then the comment

```
  // Set the inverse of the couch angle as now what moves is the room (treatment machine) around the patient,
  // so the patient support table top (couch) always stays stationary in RAS.
```

should be moved from the other files to this one.

MichaelColonel: Agreed.

MichaelColonel: Fixed.

cpinter:
Thank you very much! Since the two of you seem to agree that this works, I'm doing the merge (no time at all this week unfortunately).

Issue #262, opened 2024-12-16
SlicerRT deployment fails on Linux


cpinter:
See this forum comment: https://discourse.slicer.org/t/segment-statistics-module-on-3d-slicer-5-6-2/40714/8?u=cpinter

Basically the vtkIECTransformLogic binary is deployed to its own directory instead of `Slicer-5.7/qt-loadable-modules`.

@ferdymercury Since you worked on this on Linux, can you please take a look at this problem? Please let me know if you cannot for some reason. Thank you!

ferdymercury:
Hi Csaba, sure, I can take a look at this.

If I understand correctly, this issue must be similar to this comment by @lassoan: https://discourse.slicer.org/t/loadable-module-with-external-binary-dependencies/40667/2

But in this case, I am not sure we should add a build step that copies the .so into the qt-loadable-module folder, or whether we should instead try to set the RPATH variable and leave the .so file where it is.

Or maybe we need sth along: https://github.com/Slicer/Slicer/blob/f711b40bb6c530f1c99937536d139fee6650dace/SuperBuild/External_VTK.cmake#L223 ?

cpinter:
Yes it is similar but the problem here is that it is deployed but not quite in the right location. I'd try to find what determines the folder where it is copied at packaging. On windows it works well I think (wouldn't swear on my life, I'll check) so there may be an `#if (WIN32)` somewhere with an `install` command. I will take a look too if you cannot find it easily. Thanks a lot!

ferdymercury:
If I modify by hand the ExternalProjectAdd command
from: `-DCMAKE_LIBRARY_OUTPUT_DIRECTORY:PATH=${CMAKE_BINARY_DIR}/${Slicer_THIRDPARTY_LIB_DIR}`
to: `-DCMAKE_LIBRARY_OUTPUT_DIRECTORY:PATH=${CMAKE_BINARY_DIR}/inner-build/lib/Slicer-5.7/qt-loadable-modules`

then the .so file is output where it's indicated, but linking later on fails:

```
Linking CXX shared library ../../lib/Slicer-5.7/qt-loadable-modules/libvtkSlicerBeamsModuleLogic.so
/usr/bin/ld: cannot find /opt/SlicerRT_bld/lib/Slicer-5.7/libvtkIECTransformLogic.so: No such file or directory
```

[...]

PR #263, opened 2025-01-14
COMP: Fix Linux deployment of IEC logic library

markgerencser:
On Linux the IEC transform logic binary was deployed to it's own directory instead of Slicer-5.7/qt-loadable-modules. With this update the packaging works correctly and the IEC logic is deployed in the qt-loadable-modules folder.

Re #262

### DIFF
--- a/SuperBuild/External_vtkIECTransformLogic.cmake
+++ b/SuperBuild/External_vtkIECTransformLogic.cmake
@@ -71,7 +71,7 @@ if(NOT DEFINED ${proj}_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
       -DCMAKE_LIBRARY_OUTPUT_DIRECTORY:PATH=${CMAKE_BINARY_DIR}/${Slicer_THIRDPARTY_LIB_DIR}
       -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY:PATH=${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}
       # Install directories
-      # NA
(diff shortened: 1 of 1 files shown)


SuperBuild/External_vtkIECTransformLogic.cmake
```diff
@@ -71,7 +71,7 @@ if(NOT DEFINED ${proj}_DIR AND NOT ${CMAKE_PROJECT_NAME}_USE_SYSTEM_${proj})
       -DCMAKE_LIBRARY_OUTPUT_DIRECTORY:PATH=${CMAKE_BINARY_DIR}/${Slicer_THIRDPARTY_LIB_DIR}
       -DCMAKE_ARCHIVE_OUTPUT_DIRECTORY:PATH=${CMAKE_ARCHIVE_OUTPUT_DIRECTORY}
       # Install directories
-      # NA
+      -DvtkIECTransformLogic_INSTALL_LIB_DIR:PATH=${Slicer_QTLOADABLEMODULES_LIB_DIR}
```
lassoan: Third-party shared libraries should not be installed in a module folder. It may work on some operating systems, but probably not on all.

As I wrote in the referenced issue, please follow a working example, such as the SlicerOpenIGTLink extension:

https://github.com/openigtlink/SlicerOpenIGTLink/blob/6fbdadf16d6ccee8e840d9d408422bec4c95e867/SuperBuild/External_OpenIGTLink.cmake#L73

lassoan:
Thank you, this looks good now. To make a decision on whether it is ready to be merged, it would be good to know what OS did you test on and if you tested both the build and install tree.

cpinter:
He tried on Windows and Linux, by installing the extension from the generated zip. I'm merging and we can see the dashboard tomorrow.

PR #267, opened 2025-04-01
FIX: buffer overflow in integer when setting to 1e10 and enhance documentation

ferdymercury:
@cpinter It caused the value to be randomly cut to int (MAX_INT<1e10) (undefined behavior, platform dependent).

Convert underlying data type to long to avoid this truncation. Cast also multiplication to long, since int ^ 2 can easily surpass the limits of int.


cpinter:
Sorry for the long wait. Looks good, but please stick to the Slicer style commit prefixes. For documentation only changes use `DOC`, for fixes `BUG`, for typos `STYLE` (see [here](https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages)). Thanks a lot!

ferdymercury:
Thanks Csaba for the review! I reworded as suggested.

PR #268, opened 2025-04-01
ENH: Let user turn ON/OFF collision detection

ferdymercury:
@cpinter: In many laptops, the algorithm eats up a lot of RAM and freezes the computer for some seconds. So disable it by default, and only turn it on actively by the user, so that he/she sees why this is happening. Also, this change is fundamental since the collision detection is, at the moment, freezing the GUI rather than running in the background.

In some cases, we do not want to calculate collisions, just to play around and show functionality as well as quickly visually see where there might be collisions. Once beam angle is decided, we can activate numerical collision detection, and get more accurate results.

Thanks in advance for the review!

### DIFF
--- a/RoomsEyeView/Logic/vtkMRMLRoomsEyeViewNode.cxx
+++ b/RoomsEyeView/Logic/vtkMRMLRoomsEyeViewNode.cxx
@@ -72,7 +72,7 @@ vtkMRMLNodeNewMacro(vtkMRMLRoomsEyeViewNode);
 vtkMRMLRoomsEyeViewNode::vtkMRMLRoomsEyeViewNode()
   : PatientBodySegmentID(nullptr)
   , TreatmentMachineDescriptorFilePath(nullptr)
-  , CollisionDetectionEnabled(false)
+  , CollisionDetectionEnabled(true)
   , GantryRotationAngle(0.0)
   , CollimatorRotationAngle(0.0)
   , ImagingPanelMovement(-68.50)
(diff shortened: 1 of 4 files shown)


cpinter:
Sorry @ferdymercury another PR I have forgotten. It looks good to me, but not sure if we should make no collision detection the default. I'd surely need to modify some code to follow this change, and I think it will surprise others too. Do you have a strong point for keeping it off by default?

ferdymercury:
> Do you have a strong point for keeping it off by default?

No, no worries at all. As long as it's configurable, it's fine :)

RoomsEyeView/Logic/vtkMRMLRoomsEyeViewNode.cxx
```diff
@@ -72,7 +72,7 @@ vtkMRMLNodeNewMacro(vtkMRMLRoomsEyeViewNode);
 vtkMRMLRoomsEyeViewNode::vtkMRMLRoomsEyeViewNode()
   : PatientBodySegmentID(nullptr)
   , TreatmentMachineDescriptorFilePath(nullptr)
-  , CollisionDetectionEnabled(true)
+  , CollisionDetectionEnabled(false)
```
ferdymercury:
```suggestion
  , CollisionDetectionEnabled(true)
```

RoomsEyeView/Resources/UI/qSlicerRoomsEyeViewModule.ui
```diff
@@ -258,7 +258,7 @@
       <item row="2" column="0" colspan="3">
        <widget class="QLabel" name="CollisionDetectionStatusLabel">
         <property name="text">
-         <string>No collisions detected.</string>
+         <string>Collision computation disabled</string>
```
ferdymercury:
```suggestion
         <string>Collision computation enabled</string>
```

RoomsEyeView/Resources/UI/qSlicerRoomsEyeViewModule.ui
```diff
@@ -278,14 +278,21 @@
         </property>
        </spacer>
       </item>
-      <item row="0" column="0">
+      <item row="0" column="0" colspan="2">
+       <widget class="QCheckBox" name="ActivateCollisionComputationCheckBox">
+        <property name="text">
+         <string>Enable computation</string>
+        </property>
```
ferdymercury, line 285:
```suggestion
        </property>
        <property name="checked">
         <bool>true</bool>
        </property>
```

cpinter:
Thank you very much!

Issue #273, opened 2025-06-20
RTplan loading: unintuitive how to change energy layers (Sequence Browser Toolbar should be auto-enabled) and how to know energy and sigmaXY


ferdymercury:
@cpinter
I am trying to load an RTplan with 1000 proton pencil beam spots. Every spot belongs to a different energy layer, there are 10 layers, so 100 spots per layer.
Currently, SlicerRT only visualizes the first energy layer and one can not visualize the other 900 spots, there is no way to change the energy via dropdown menu (or at least I can't find it).

Also, the source distance is 2m but SlicerRT shows it as 5m.

I can send you a confidential demonstrator file if needed.
Thanks in advance!

MichaelColonel:
May I have a look at these files? Without DICOM Conformance Statement from a particular manufacturer, the correct support of ion plans is impossible.

ferdymercury:
Sure, find attached.
[watercube.zip](https://github.com/user-attachments/files/20852384/watercube.zip)
Thanks in advance.

I get:
```
Warning: In vtkMRMLRTIonBeamNode.cxx, line 386
vtkMRMLRTIonBeamNode (0xa03b3b0): CreateBeamPolyData: Wrong number of leaf pairs in the MLC table node, beam model poly data will be created without MLC
Warning: In vtkMRMLRTIonBeamNode.cxx, line 374
vtkMRMLRTIonBeamNode (0xa03b3b0): CreateBeamPolyData: Invalid or absent table node with scan spot parameters for a node "G000"
```

And only the first spot is visible in the spot table.

I think the problem is that SlicerRT does not handle well the case where ScanMode DICOM tag is 'MODULATED' (pencil-beam scanning) instead of passive scattering.

Here, every spot has an energy, a sigmaX and a sigmaY, that should be shown in the table, too.

MichaelColonel:
In file above all 58 control points contain ScanSponPositionMap with only one point (NumberOfScanSpotPositions) with position (0,0) and scanning spot size 6/6, it's not a array with different (x,y) pairs. What kind of information do you want to display or visualize, and how do you want to use kinetic energy value?

ferdymercury:
Well, that was just an example. I have other examples where there are, for each different energy layer, about 200 NumberOfScanSpotPositions with different x,y values.

The table should contain for each spot:
x y Energy sigmaX sigmaY
Or alternatively, that there is one separate table for each energy, containing x y sigmaX sigmaY

[...]

Issue #256, opened 2024-10-15
Proton dose engine calculates dose with incorrect source position


cpinter:
The source position we set to the Plastimatch proton dose engine is correct, but the output dose looks like if the SAD were inverted (the source on the other side of the isocenter).

![image](https://github.com/user-attachments/assets/ae059e3a-0f7b-488a-81c0-febed9cc6521)

It worked the same way in Slicer 5.2.2 so the problem exists at least for a few years.

Issue #237, opened 2023-10-25
TROTS dataset


ferdymercury:
Feature suggestion @cpinter :

In "main" 3D Slicer, you can download several patient-examples thanks to the "Download Sample" plugin.

It would be nice if SlicerRt had an equivalent module that would download this dataset:
https://sebastiaanbreedveld.nl/trots/

There are zenodo permalinks to it, and h5, .dcm versions as well as conversion scripts.

https://github.com/SebastiaanBreedveld/TROTS/blob/master/Scripts%20DICOM/writeTROTSToDICOM.py

cpinter:
The SampleData module in Slicer supports extensions. SlicerRT already adds a dataset to this module (see before and after installing SlicerRT).

If it's not important to load this sample data from DICOM, then I can add it the same way as the "TinyRTPatient" that is already added.

ferdymercury:
> If it's not important to load this sample data from DICOM, then I can add it the same way as the "TinyRTPatient" that is already added.

Sounds perfect

cpinter:
I just looked at the dataset. It consists of many patients (145 it seems). The SampleData module is to load a single volume, which is extended with the segmentation in the case of the above mentioned TinyRtPatient dataset. I cannot add the 145 patients to a single "sample dataset", it would load an unreasonable amount of data. I also would like to avoid adding them one button each. All I can think of as reasonable is to choose a few patients and add those as a button each. Did you have something in mind?

ferdymercury:
Thanks for the reply! 

 was thinking more of "6 buttons".

- Prostate_CK contains 30
- Prostate_VMAT contains 30
- Head-and-Neck 15
- Protons 20
- Liver set contains 10
- Prostate_BT set 25

If 25 or 30 is still too much, then maybe adding a popup asking how many patients of the group it should load (which by default should only download one)?

cpinter:
The SampleData module is in Slicer core and adding new options is not that simple. This would mean adding the feature just for the purposes of one extension, which would be an exception unless there is wider use of this.

So you were thinking that clicking a button would load up to 30 patients and use all of them in the same Slicer run? Can you please let me know how such a Slicer run with 30 patients loaded would look like?

@lassoan Do you have any thoughts on this so far? Thank you!

ferdymercury:
Mmm I was thinking more that it would download them to the local DICOM database, and let you choose which patient to open. Not opening them all at the same time. (Sorry for being unclear hehe)

[...]

PR #274, opened 2025-06-25
ENH: Improvements in RTIonPlan loading logic

MichaelColonel:
Possible solution for Issue #273.

DicomRtImportExportModuleLogic:
Energy value is saved within beam node
IonPlan flag is saved within plan node

BeamParametersTabWidget:
VSAD is correctly displayed in BeamParametersTab

### DIFF
--- a/Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
+++ b/Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
@@ -32,6 +32,7 @@ Ontario with funds provided by the Ontario Ministry of Health and Long-Term Care
 
 #include "vtkMRMLRTBeamNode.h"
 #include "vtkMRMLRTIonBeamNode.h"
+#include "vtkMRMLRTIonRangeShifterNode.h"
 #include "vtkMRMLRTPlanNode.h"
 
 // MRML includes
@@ -52,6 +53,7 @@ Ontario with funds provided by the Ontario Ministry of Health and Long-Term Care
 // Qt includes
 #include <QDebug>
 #include <QLineEdit>
+#include <QSignalBlocker>
 
 //----------------------------------------------------------------------------
 const char* qMRMLBeamParametersTabWidget::BEAM_PARAMETER_NODE_ATTRIBUTE_PROPERTY = "BeamParameterNodeAttribute";
@@ -113,6 +115,9 @@ void qMRMLBeamParametersTabWidgetPrivate::init()
   QObject::connect( this->pushButton_GenerateMLCBoundary, SIGNAL(clicked()), q, SLOT(generateMLCboundaryClicked()) );
   QObject::connect( this->pushButton_UpdateMLCBoundary, SIGNAL(clicked()), q, SLOT(updateMLCboundaryClicked()) );
   QObject::connect( this->pushButton_CalculateMLCPosition, SIGNAL(clicked()), q, SLOT(calculateMLCPositionClicked()) );
+
+  // Scan spot weights table page
+  QObject::connect( this->MRMLTableView_ScanSpotWeight, SIGNAL(selectionChanged()), q, SLOT(onScanSpotWeightSelectionChanged()) );
 
   // Load MLC position calculation logic
   qSlicerCoreApplication* app = qSlicerCoreApplication::application();
@@ -191,25 +196,54 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
   d->doubleSpinBox_SAD->setValue(d->BeamNode->GetSAD());
   d->RangeWidget_XJawsPosition->setValues(d->BeamNode->GetX1Jaw(), d->BeamNode->GetX2Jaw());
   d->RangeWidget_YJawsPosition->setValues(d->BeamNode->GetY1Jaw(), d->BeamNode->GetY2Jaw());
-  d->SliderWidget_CollimatorAngle->setValue(d->BeamNode->GetCollimatorAngle());
-  d->SliderWidget_GantryAngle->blockSignals(true);
-  d->SliderWidget_GantryAngle->setValue(d->BeamNode->GetGantryAngle());
-  d->SliderWidget_CouchAngle->setValue(d->BeamNode->GetCouchAngle());
-  d->SliderWidget_GantryAngle->blockSignals(false);
 
-  d->MRMLNodeComboBox_MLCBoundaryAndPositionTable->setMRMLScene(d->BeamNode->GetScene());
+  // prevent transform update
+  {
+    const QSignalBlocker blocker(d->SliderWidget_CollimatorAngle);
+    d->SliderWidget_CollimatorAngle->setValue(d->BeamNode->GetCollimatorAngle());
+  }
+  {
+    const QSignalBlocker blocker(d->SliderWidget_GantryAngle);
+    d->SliderWidget_GantryAngle->setValue(d->BeamNode->GetGantryAngle());
+  }
+  {
+    const QSignalBlocker blocker(d->SliderWidget_CouchAngle);
+    d->SliderWidget_CouchAngle->setValue(d->BeamNode->GetCouchAngle());
+  }
+
+  // Set control point isocenter (if absent set widget disabled and use plan's isocenter)
+  double isocenter[3] = {};
+  bool planIsocenterFlag = false;
+  if (d->BeamNode->GetIsocenterPositionFlag())
+  {
+    d->BeamNode->GetIsocenterPosition(isocenter);
+    d->CoordinatesWidget_Isocenter->setEnabled(true);
+  }
+  else
+  {
+    planIsocenterFlag = d->BeamNode->GetPlanIsocenterPosition(isocenter);
+  }
+  d->CoordinatesWidget_Isocenter->setCoordinates(isocenter);
+  if (planIsocenterFlag)
+  {
+    d->CoordinatesWidget_Isocenter->setEnabled(false);
+  }
 
   // Check for MLC table and enable combo box
   if (vtkMRMLTableNode* mlcTable = d->BeamNode->GetMultiLeafCollimatorTableNode())
   {
+    d->MRMLNodeComboBox_MLCBoundaryAndPositionTable->setMRMLScene(mlcTable->GetScene());
     d->MRMLNodeComboBox_MLCBoundaryAndPositionTable->setCurrentNode(mlcTable);
+    d->MRMLNodeComboBox_MLCBoundaryAndPositionTable->setEnabled(true);
     d->pushButton_UpdateMLCBoundary->setEnabled(true);
+    d->doubleSpinBox_DistanceMLC->setEnabled(true);
   }
   else
   {
     d->MRMLNodeComboBox_MLCBoundaryAndPositionTable->setCurrentNode(nullptr);
     d->MRMLNodeComboBox_MLCBoundaryAndPositionTable->setEnabled(false);
     d->pushButton_UpdateMLCBoundary->setEnabled(false);
+    d->doubleSpinBox_DistanceMLC->setEnabled(false);
   }
 
   // Update engine-specific values
@@ -241,34 +275,30 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
         QCheckBox* checkBox = qobject_cast<QCheckBox*>(currentParameterFieldWidget);
         if (lineEdit)
         {
-          lineEdit->blockSignals(true);
+          const QSignalBlocker blocker(lineEdit);
           lineEdit->setText(parameterValue);
-          lineEdit->blockSignals(false);
         }
         else if (slider)
         {
-          slider->blockSignals(true);
+          const QSignalBlocker blocker(slider);
           slider->setValue(parameterValue.toDouble());
-          slider->blockSignals(false);
         }
         else if (spinBox)
         {
-          spinBox->blockSignals(true);
+          const QSignalBlocker blocker(spinBox);
           spinBox->setValue(parameterValue.toDouble());
-          spinBox->blockSignals(false);
         }
         else if (comboBox)
         {
-          comboBox->blockSignals(true);
+          const QSignalBlocker blocker(comboBox);
           comboBox->setCurrentIndex(parameterValue.toInt());
-          comboBox->blockSignals(false);
         }
         else if (checkBox)
         {
-          checkBox->blockSignals(true);
-          checkBox->setChecked(QVariant(parameterValue).toBool());
-          checkBox->blockSignals(false);
-
+          {
+            const QSignalBlocker blocker(checkBox);
+            checkBox->setChecked(QVariant(parameterValue).toBool());
+          }
           // Enable/disable dependent parameters
           this->updateDependentParameterWidgetsForCheckbox(checkBox);
         }
@@ -297,6 +327,15 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
     // show widgets for rt beam
     d->doubleSpinBox_SAD->show();
     d->label_SAD->show();
+    {
+      const QSignalBlocker blocker(d->doubleSpinBox_SAD);
+      d->doubleSpinBox_SAD->setValue(d->BeamNode->GetSAD());
+    }
+
+    d->CollapsibleButton_RangeShifterParameters->setCollapsed(true);
+    d->CollapsibleButton_RangeShifterParameters->setEnabled(false);
+    d->CollapsibleButton_SnoutParameters->setCollapsed(true);
+    d->CollapsibleButton_SnoutParameters->setEnabled(false);
   }
   else if (ionBeamNode)
   {
@@ -307,7 +346,7 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
     // Check for ScanSpot table and enable combo box
     if (vtkMRMLTableNode* scanspotTable = ionBeamNode->GetScanSpotTableNode())
     {
-      d->MRMLNodeComboBox_ScanSpotParametersTable->setMRMLScene(d->BeamNode->GetScene());
+      d->MRMLNodeComboBox_ScanSpotParametersTable->setMRMLScene(scanspotTable->GetScene());
       d->MRMLNodeComboBox_ScanSpotParametersTable->setCurrentNode(scanspotTable);
     }
     else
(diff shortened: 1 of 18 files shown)


ferdymercury:
Thanks a lot, also some more suggestions:

- Could the Beam Isocenter coordinates be added to the Geometry of the BeamParameters dialog ?
- For Ion Plans, the RangeShifter details (Name, SnoutPosition or BlockTrayDistance) could appear in a "tab" of the Beam Parameters dialog.
- And probably Isocenter-to-MLC, JawPositionsXY should be grayed out if MLC is none.

ferdymercury:
> VSAD is correctly displayed in BeamParametersTab

Maybe related: https://github.com/SlicerRt/SlicerRT/issues/256

MichaelColonel:
Some improvements in ScanSpotMap visualization.

MichaelColonel:
> * Could the Beam Isocenter coordinates be added to the Geometry of the BeamParameters dialog ?

Do you need this information as read only, or you want to change the isocenter position?

> For Ion Plans, the RangeShifter details (Name, SnoutPosition or BlockTrayDistance) could appear in a "tab" of the Beam Parameters dialog.

Only as a text messages (labels), range shifters are also heavily depends on physical realization.

ferdymercury:
As readonly information, thanks.
Yes, as text labels.
Thanks again

MichaelColonel:
There is no "Range Shifter Name" tag within "Range Shifter Sequence" , only  "Range Shifter ID".
By BlockTrayDistance do you mean "Isocenter to Block Tray Distance" tag within "Ion Block Sequence"?

ferdymercury:
> There is no "Range Shifter Name" tag within "Range Shifter Sequence" , only "Range Shifter ID". By BlockTrayDistance do you mean "Isocenter to Block Tray Distance" tag within "Ion Block Sequence"?

Sorry, yes, I was thinking of RangeShifterID and RangeShifterType

Yep, Iso-toBlockTray-Distance.

ferdymercury:
Btw, I just noticed that the Range-Shifter might change during the beam, see https://dicom.innolitics.com/ciods/rt-ion-plan/rt-ion-beams/300a03a2/300a03a8/300a0360
and it also has a RS to Isocenter distance.

MichaelColonel:
Do you have a  DICOM Conformance Statement for the machine? It's no use to write a universal code without knowing a particular implementation.

Each range shifter can be implemented as a separate node with it's own model. But correct visualization of all beam components (ion blocks, wedges, range shifters, range compensators, boluses) is almost impossible.

ferdymercury:
> DICOM Conformance Statement for the machine

Yes, see https://helpcenter.iba-dosimetry.com/hc/en-us/article_attachments/10381360329500

Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
```diff
@@ -297,6 +318,9 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
     // show widgets for rt beam
     d->doubleSpinBox_SAD->show();
     d->label_SAD->show();
+    d->doubleSpinBox_SAD->blockSignals(true);
```
cpinter: Can you please use `QSignalBlocker`? It is a more convenient way to do the same thing (normally we'd need to save the return value of the first call (`wasBlocked = `) and use it in the second call, but with signal blocker it is automatic)

cpinter:
Thank you very much! I just have the one minor comment, then we can integrate.

MichaelColonel:
I've added range shifter and snout info from control points in BeamParametersWidget (temporary in Geometry tab, but it can be moved elsewhere).

I've only test dataset for RTIonPlan with only two control point (static beam), and i can't guarantee that it works for a dynamic beam with multiple control points.
image

ferdymercury:
> I've only test dataset for RTIonPlan with only two control point (static beam), and i can't guarantee that it works for a dynamic beam with multiple control points.

In case you want to try it out with dynamic beams, find [here](https://consigna.uv.es/?recoge:ca:3045_9241_2262_7679) the TROTS dataset with 20 proton patients, many of them with a RangeShifter changing within a beam (IN/OUT). It loads fine directly from 3D Slicer DICOM import.

And thanks a lot again!

Another thing that I was thinking is that, for proton therapy, unlike radiotherapy, it does not make sense to paint the beams in green to be symmetrical with respect to the patient, since the protons stop. It would be nice if the 3D green rectangles representing the beam started at the virtual source axis distance, but then stopped / were clipped at the end of the patient bounding box. Or 50 cm away, not 2m away. Otherwise it's harder to know the direction where they are coming from.

MichaelColonel:
Most of the issues were fixed.

> Or 50 cm away, not 2m away. Otherwise it's harder to know the direction where they are coming from.

I've set 50 cm after isocenter.

It's ready for a final review.

ferdymercury:
Awesome.
Thanks a lot!!

This is maybe too much to ask, but it could be nice that, if one selects with the mouse a specific row (or several of them, by drag-and-hold) of the ScanSpotPositionMap table, the corresponding X-Y position (ellipse/cylinder) gets highlighted in the 3D view (or changes color or whatever).

MichaelColonel:
How do you suggest to invoke custom event for vtkTableNode with selected table rows as data to update RTIonNode?

lassoan:
I think there are some extensions (maybe SlicerProstate) that observe table widget selection changes. It should work by connecting a slot to the "selection changed" signal of the "selection model" of the table widget.

ferdymercury:
Maybe this helps: https://discourse.slicer.org/t/get-active-cell-from-loaded-table/10882

MichaelColonel:
> This is maybe too much to ask, but it could be nice that, if one selects with the mouse a specific row (or several of them, by drag-and-hold) of the ScanSpotPositionMap table, the corresponding X-Y position (ellipse/cylinder) gets highlighted in the 3D view (or changes color or whatever).

As an example: [ScanSpot selection for highlighting](https://rutube.ru/video/private/cbd1270a80de921f9cf99b1a20c608cc/?p=4GxoQfaUQivPnQkgTf-h5A)

MichaelColonel:
Short video: [Interactive highlighting](https://rutube.ru/video/private/bb06614de04e17a3ea5b550e2cad6e8f/?p=TI_vsJgGaKpDEwOLQ6QPAQ&t=21&r=plwd)

ferdymercury:
Just awesome! You're the best. Thanks a lot

ferdymercury:
One additional observation for the future is the following:

When one loads an RTionplan, all beams are shown in green with a size corresponding ti the first item of the ion-control-point-sequence.
It would be beneficial to have a checkbox to visualize the aggregate XY contributions from the different items in the sequence (ideally just summing the even control point indices, since the other are zero). To get an overview of the full beam, not just of a given index of the sequence.

ferdymercury:
@MichaelColonel it could be interesting if this info from the rtplan.dcm is shown in a view-only table directly in SlicerRt

image

MichaelColonel:
What's wrong with the DICOM viewer in Slicer?

ferdymercury:
Hmm since this contains important information about the plan, it's relevant to show it in a table, rather than having do dive through DICOM tags and unfolding items of the sequence one by one, etc in the Slicer-DICOM-viewer.

MichaelColonel:
It's not a part of DICOM Conformance Statement you sent before. What tags do you want to be in the table? There are about 20 tags total within Dose Reference Sequence Item, and displaying all of them are not very functional.

ferdymercury:
> It's not a part of DICOM Conformance Statement you sent before. What tags do you want to be in the table? There are about 20 tags total within Dose Reference Sequence Item, and displaying all of them are not very functional.

Thanks. I'd be interested in MinDose, PrescriptionDose, MaxDose, Weight, Description, ReferenceType and UnderdoseVolume. MinDose would be TargetMinDose or OrganMinDose depending on if RefType is Target or Organ. Prescription and Underdose only if is Target.

ferdymercury:
> It's not a part of DICOM Conformance Statement you sent before.

The one I sent before was for the machine (accelerator) manufacturer.

Additionally, there is the conformance statement from the TPS vendor, see https://www.raysearchlabs.com/siteassets/raystation-landing-page/dicom-conformance-statements/raystation-pdfs/rsl-d-rs-11a-dcs-en-1.0-2021-05-07-raystation-11a-dicom-conformance-statement.pdf
see section 7.1.5.6

DicomRtImportExport/Logic/vtkSlicerDicomRtImportExportModuleLogic.cxx
```diff
@@ -775,6 +775,41 @@ bool vtkSlicerDicomRtImportExportModuleLogic::vtkInternal::LoadRtPlan(vtkSlicerD
...
+    std::string tableName = name.str();
+    doseReferencesTableNode->SetName(tableName.c_str());
+    scene->AddNode(doseReferencesTableNode);
+
+    doseReferencesTableNode->SetUseColumnTitleAsColumnHeader(true);
+    doseReferencesTableNode->SetColumnDescription("Number", "Dose Reference Number");
+    doseReferencesTableNode->SetColumnDescription("Structure Type", "Dose Reference Structure Type");
+    doseReferencesTableNode->SetColumnDescription("Minimum Dose", "Target Minimum Dose");
+    doseReferencesTableNode->SetColumnDescription("Prescription Dose", "Target Prescription Dose");
+    doseReferencesTableNode->SetColumnDescription("Maximum Dose", "Target Maximum Dose");
```
ferdymercury:
```suggestion
    doseReferencesTableNode->SetColumnDescription("Minimum Dose", "Target/Organ Minimum Dose");
    doseReferencesTableNode->SetColumnDescription("Prescription Dose", "Target Prescription Dose");
    doseReferencesTableNode->SetColumnDescription("Maximum Dose", "Target/Organ Maximum Dose");

```

MichaelColonel:
I've added initial support for DoseReferenceSequence.
Table node is added to plan node.

Name template: PlanNodeName :  Dose reference Table [Prescription Description].

Test it.

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -354,6 +358,31 @@ class vtkSlicerDicomRtReader::vtkInternal
...
+  class DoseReferenceEntry
+  {
+  public:
+    DoseReferenceEntry()
+    {
+      Number = -1;
+      TargetMinimumDose = -1.;
+      TargetPrescriptionDose = -1.;
+      TargetMaximumDose = -1.;
+      ConstaintWeight = -1.;
```
ferdymercury:
```suggestion
      ConstraintWeight = -1.;

```

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -354,6 +358,31 @@ class vtkSlicerDicomRtReader::vtkInternal
...
+    }
+    unsigned int Number;
+    std::string StructureType;
+    double TargetMinimumDose;
+    double TargetPrescriptionDose;
+    double TargetMaximumDose;
+    double ConstaintWeight;
```
ferdymercury:
```suggestion
    double ConstraintWeight;

```

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -2156,6 +2250,81 @@ void vtkSlicerDicomRtReader::vtkInternal::LoadRTStructureSet(DcmDataset* dataset
...
+      continue;
+    }
+    DoseReferenceEntry doseReferenceEntry;
+    OFCondition cond;
+
+    Float64 weight = -1.;
+    cond = doseReferenceItem.getConstraintWeight(weight);
+    if (cond.good())
+    {
+      doseReferenceEntry.ConstaintWeight = weight;
```
ferdymercury:
```suggestion
      doseReferenceEntry.ConstraintWeight = weight;

```

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -2808,6 +2978,79 @@ int vtkSlicerDicomRtReader::GetNumberOfBeams()
...
+    table->SetNumberOfRows(this->Internal->DoseReferenceSequenceVector.size());
+    vtkIdType row = 0;
+    for(auto& doseRefEntry : this->Internal->DoseReferenceSequenceVector)
+    {
+      table->SetValue(row, 0, doseRefEntry.Number);
+      table->SetValue(row, 1, doseRefEntry.StructureType.c_str());
+      table->SetValue(row, 2, doseRefEntry.TargetMinimumDose);
+      table->SetValue(row, 3, doseRefEntry.TargetPrescriptionDose);
+      table->SetValue(row, 4, doseRefEntry.TargetMaximumDose);
+      table->SetValue(row, 5, doseRefEntry.ConstaintWeight);
```
ferdymercury:
```suggestion
      table->SetValue(row, 5, doseRefEntry.ConstraintWeight);

```

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -354,6 +358,31 @@ class vtkSlicerDicomRtReader::vtkInternal
...
+      TargetMinimumDose = -1.;
+      TargetPrescriptionDose = -1.;
+      TargetMaximumDose = -1.;
+      ConstaintWeight = -1.;
+    }
+    unsigned int Number;
+    std::string StructureType;
+    double TargetMinimumDose;
+    double TargetPrescriptionDose;
+    double TargetMaximumDose;
```
ferdymercury, line 380:
```suggestion
    double TargetMaximumDose;
    double OrganMaximumDose;
```

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -354,6 +358,31 @@ class vtkSlicerDicomRtReader::vtkInternal
...
+  /// Structure storing an RT Dose Prescription, Dose reference
+  class DoseReferenceEntry
+  {
+  public:
+    DoseReferenceEntry()
+    {
+      Number = -1;
+      TargetMinimumDose = -1.;
+      TargetPrescriptionDose = -1.;
+      TargetMaximumDose = -1.;
```
ferdymercury, line 370:
```suggestion
      TargetMaximumDose = -1.;
      OrganMaximumDose = -1.;
```

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -2156,6 +2250,81 @@ void vtkSlicerDicomRtReader::vtkInternal::LoadRTStructureSet(DcmDataset* dataset
...
+    cond = doseReferenceItem.getTargetPrescriptionDose(dose);
+    if (cond.good())
+    {
+      doseReferenceEntry.TargetPrescriptionDose = dose;
+    }
+    cond = doseReferenceItem.getTargetMaximumDose(dose);
+    if (cond.good())
+    {
+      doseReferenceEntry.TargetMaximumDose = dose;
+    }
```
ferdymercury: This only if StructureType == "TARGET"
else:
doseReferenceEntry.OrganMaximumDose = GetOrganAtRiskMaximumDose(dose)

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
```diff
@@ -2808,6 +2978,79 @@ int vtkSlicerDicomRtReader::GetNumberOfBeams()
...
+    // Column 3; PrescriptionDose
+    vtkNew<vtkDoubleArray> prescDose;
+    prescDose->SetName("Target Prescription Dose");
+    table->AddColumn(prescDose);
+
+    // Column 4; MaxDose
+    vtkNew<vtkDoubleArray> maxDose;
+    maxDose->SetName("Target Maximum Dose");
+    table->AddColumn(maxDose);
+
```
ferdymercury, line 3055: Another column for OrganAtRiskMaxDose

ferdymercury (commented):
Thanks a lot again! I will compile it this afternoon and test it!

In the meanwhile, I added a couple of suggestions:

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
MichaelColonel: StructureType? Not ReferenceType?

The standard says that StructureType: POINT, VOLUME, COORDINATES, SITE;
                                        ReferenceType: TARGET, ORGAN_AT_RISK.

MichaelColonel:
image

DicomRtImportExport/Logic/vtkSlicerDicomRtReader.cxx
ferdymercury: sorry, you're right, ReferenceType

ferdymercury:
Looks awesome, thanks!!

Some more feedback:
- Description should appear as first column
- Number Column is probably not necessary.
- Showing an empty cell instead of -1 when something is undefined would be better.
- StructureType is not too relevant and could be avoided as column.
- Maybe, ReferenceType column could be avoided too, since we already know what is target and what is organ at risk depending on if OrganAtRiskMaxDose was defined or TargetMaxDose was defined.
- Maximum and Minimum could be abbreviated as Max. and Min. for those two target column titles
- With the won space, we can maybe add two new columns: TargetUnderdose and OrganOverdose

MichaelColonel:
Empty cells don't look pleasant.
image

ferdymercury:
Leaving here a link to 20-proton-patient dataset to easen testing of this PR: https://consigna.uv.es/?recoge:ca:3245_9718_9829_7607

lassoan:
Would it make sense to add it to https://github.com/SlicerRt/SlicerRtData?

ferdymercury:
> Would it make sense to add it to https://github.com/SlicerRt/SlicerRtData?

Could be an option, yes, related discussion here: https://github.com/SlicerRt/SlicerRT/issues/237
either the data, or the Python script directly that downloads the data from zenodo and converts them to DICOM.

MichaelColonel:
> Leaving here a link to 20-proton-patient dataset to easen testing of this PR: https://consigna.uv.es/?recoge:ca:3245_9718_9829_7607

For some sets "Isocenter To Range Shifter Distance" tag (300A,0364) has meaningless value (too short), so range shifter block model is within human body.

ferdymercury:
> For some sets "Isocenter To Range Shifter Distance" tag (300A,0364) has meaningless value (too short), so range shifter block model is within human body.

Thanks for checking. That tag is optional, and is not defined in any of the 20 cases.
See https://github.com/SebastiaanBreedveld/TROTS/blob/master/Scripts%20DICOM/writeTROTSToDICOM.py
There is no IsocenterToRangeShifterDistance defined there.

Could it be that it's being filled with a garbage value from memory during loading?

I guess that, if it's not defined, it should be set at say 50cm from ISO or sth like that.

ferdymercury:
(Btw there's a merge conflict)

MichaelColonel:
> (Btw there's a merge conflict)

That was a wrong branch with wrong commits.
I've set a default distance for isocenter to range shifter to 50 cm.

cpinter:
@ferdymercury Do you agree with this change as is currently? I won't be able to review this in detail for at least a week, but if you consider it OK we can merge it. Thank you @MichaelColonel and @ferdymercury !

ferdymercury:
Hi Csaba, thanks for the feedback. I could not test this PR myself because of https://discourse.slicer.org/t/slicerrt-built-from-source-cannot-start/43792

But I agree with the changes!

cpinter:
Of course. Let me think. It would be great if you could test it first.

ferdymercury:
I was able now to build and test the PR, it looks great to me, I approve, thanks a lot again Michael!

cpinter:
Thank you very much both!

PR #275, opened 2025-07-14
Inverse Planning Support and pyRadPlan Integration in SlicerRT's External Beam Planning Module

LiBu981:
# Summary
This pull request extends SlicerRT's **External Beam Planning (EBP)**  module to support a more modular architecture for treatment planning by separating _dose calculation_ and _optimization_ for _inverse planning_, while focusing on integrating the treatment planning toolkit **pyRadPlan**.

# Key Changes

### **Separation of Dose Calculation and Optimization**

- Code under `ExternalBeamPlanning\Widgets\Python` has been restructured into
     - `DoseEngines`
     - `PlanOptimizers`
   
- When _"Inverse Treatment Planning?"_ is selected, _dose calculation_ and _plan optimization_ are handled as two distinct steps.
   
- This separation allows different engines to be used for each task independently.

### **Inverse Planning**

- Each dose engine reports whether it supports inverse planning (`isInverse` flag).
   
- If supported and _"Inverse Treatment Planning?"_ is selected:
     - The dose engine calculates a _dose influence matrix_ (as sparse matrix) for each beam.
     - The matrix may be stored in the corresponding `beamNode`.
   
- The plan optimizer:
     - Performs optimization based on the `planNode` and selected `objectives`. 
     - Can access stored _dose influence matrices_ of the `beamNodes` under the `planNode` (as sparse matrix in C++ or as field data in python).
     - Is required to load the final optimized dose to the `resultOptimizationVolumeNode`.
     `optimizePlanUsingOptimizer(planNode, objectives, resultOptimizationVolumeNode)`
 

### **Objectives Table**

- A new UI table (based on _qMRMLBeamsTableView_) allowing the user to define objectives for inverse planning.
  
- For each objective (each row) the user selects:
    - **_segment_** (from selected structure set) on which the objective is applied
    - **_overlap priority_** of the segment
    - **_penalty_** for the objective in the optimization
    - **_parameters_** specified to the objective function

- An `objectiveNode` is created for each row, holding...
    - names of objective function & assigned segment
    - overlap priority, penalty & parameters (as attributes)
        
   ... and is saved in the optimizer
       
- For selection _available objectives_ are defined by the optimizer and stored as `std::vector<ObjectiveStruct>`, where:
```
struct ObjectiveStruct
{
      std::string name;
      std::map<std::string, std::string> parameters;
}
```

- An _abstract objective class_ is introduced for implementing objective functions and gradients in C++. As an example, `qSlicerSquaredDeviationObjective` is provided and registered as an available objective for the _Mock Optimizer_, though it is not yet used during the optimization process.

### **Integration of pyRadPlan**

- On opening the **EBP** module, the system checks for a _pyRadPlan installation_ and prompts the user to install if needed.
    
- The following Python-based components are integrated:
     - `pyRadPlanEngine.py` --> dose engine
     - `pyRadPlanPlanOptimizer.py` --> plan optimizer
     - `PyRadPlanUtils.py` --> constructs `ct`, `cst` and `pln` data structures from the `beamNode`, `referenceVolume` and `planNode`
         (pyRadPlan applies stable validation and serialization with pydantic)

 - _Dose calculation_ (inverse planning):
     - The _radiation mode_ is selected per beam in the "pyRadPlan parameters" tab of the _Beams_ module. When _"Ion plan"_ is selected only "proton" and "carbon" are available.
     - For each beam:
          - `stf` and `dij` data structures are computed from `ct`, `cst` and `pln`.
          - The _dose influence matrix_ (`dij`) is saved in the `beamNode`, along with the corresponding _dose grid dimension_ and _spacing_.
        
- _Optimization_:
     - Loads the _dose influence matrices_ from all beams and composes the total `dij`.
     - Groups the _objectives_ by segment and applies them to the `cst`.
     - Performs _fluence optimization_ on the total composed `dij`.
     - _Beam-wise dose contributions_ are visualized as `ScalarVolumeNodes`.

# Known Limitations / Future Works

- **Stability**: The system occasionally crashes when creating a new `planNode` on startup - this requires further investigation and handling.
- **pyRadPlan Integration**: As **pyRadPlan** is still under development and only a subset of its capabilities is currently exposed in this pull request, future improvements will focus on deeper integration and further UI extensions to support additional features.
-  **Constraints Management**: Currently, only objectives are supported. A similar interface and data handling structure for defining and managing _constraints_ should be implemented in future work.
- **Forward Planning Support**: The current architectural modifications focus primarily on inverse planning workflows. Future works may also aim to incorporate forward planning procedures.
- **UI Enhancements**: The interface for defining and managing objectives, selecting engines, and visualizing results may benefit from refinement to improve clarity and usability.


ferdymercury:
Congratulations for this awesome work!!!

I have one question:

> ### **Objectives Table**
> 
>     * A new UI table (based on _qMRMLBeamsTableView_) allowing the user to define objectives for inverse planning.

Shouldn't there be a separate table and "C++ struct" for constraints and another for objectives?

LiBu981:
Yes, thank you for pointing this out. I forgot to mention this in the "Future Works". Currently pyRadPlan is not handling constraints, so we set the focus on the objectives for now. But yes, there should be a similar structure/table for constraints in the future.

ferdymercury:
Thanks for clarifying :)

One observation, we could maybe consider to have a dictionary of predefined standard optimization functions, stored in a dict on the C++ side, and shown as a dropdown menu in the GUI, following a bit https://youtu.be/FIWYQRacRDI?feature=shared&t=326
Then, an extra option "Other" where the user can specify more-custom optimization functions.

Other suggestions for "future work", would be to have a "Robust" checkbox (even if not support at this stage, it would be good to leave room for it in the GUI); and to include a optimization-progress feedback in the GUI / progress bar or so.

Another comment: Weight instead of Penalty might be more physician-friendly as label.

cpinter:
Just letting everyone know that I'm back from vacation, and I'll look at this ASAP! This is very exciting, thank you so much for the hard work!

ferdymercury:
This 20-proton-patients dataset might be helpful for stress-testing this PR: https://consigna.uv.es/?recoge:ca:3245_9718_9829_7607

LiBu981:
Hi everyone, sorry for the late reply. I also have more time now and will be available to actively work on the pull request in the upcoming weeks.

To comment on the optimization/objective functions:
Currently the available objective functions are defined by each optimizer ("setAvailableObjectives"). Stored are only objective name and its parameters. Would it maybe make sense to have a fixed set of C++ based functions in the dropdown menu, and then add the specific functions for each optimizer (instead of an "Other" option), e.g. with **'objectiveName'_myOptimizer**? Or could this become to confusing with too many objectives?
The inclusion of a fixed set should easily be done by setting the "availableObjectives"  in the abstract optimizer.

I'll add the robust checkbox and try including an optimization progress bar.

And yes, I would greatly appreciate feedback on any naming conventions. I mostly oriented myself on pyRadPlan's conventions here, which might not always be transferable.

ferdymercury:
Thanks a lot!

Regarding names, etc. I leave here a link https://consigna.uv.es/?recoge:ca:7746_1317_7296_5505 with screenshots of GUIs to get inspiration from. Maybe it's good if we can make naming match the most commonly used conventions.

Regarding checkbox: would it also make sense to create a "Mean" checkbox? Or do we prefer having separate "MaxDose" and "MeanMaxDose" objectives?

> with **'objectiveName'_myOptimizer**?

Or maybe instead, one has to choose first the optimizer, and then the content of the dropdown menu is dynamically adjusted?

cpinter:
@LiBu981 I did a quick (2-hour) review round of this PR and it looks quite good! Thank you very much for the work you've done on this!

I only have one concrete comment currently: I'd rename the `prepareRTDataset.py` file so that it becomes obvious that it is a utility for the PyRadPlan features. Maybe simple rename to `PyRadPlanUtils.py` where you can later include other PyRadPlan specific functions that are not only for preparing datasets?

Also a question: in your branch does forward planning still work as before?

I'll continue the review in the following days. After that maybe we could have a meeting, even with @ferdymercury or @wahln to discuss some details, facilitating the review process?

wahln:
Maybe a (not so) short comment from my side on the design choice of letting the optimizer define the objectives that can be used itself.

We first thought about providing a general objective & constraint interface (i.e. similar to the dose engines or the optimizer, objectives and, in the future, constraints can be defined in Python or C++). However, we were concerned about the computational overhead of this, as then data would need to be transfered during each iteration of the optimization if you mix a python optimizer with a C++ objective. 

Sometimes RT optimizers also have dedicated implementations (they need 2nd order derivatives, or use some tricks in evaluating linear constraints, and so on) that would be difficult to represent in a general interface. 

Also, what if you have a very specialized optimization routine doing something really different, like working with dose rate, biological modeling, or LET on top - this would also already create a headache. At some point you would end up with defining so many flags / weird class hierarchy situtations to make sure an optimizer understands capabilities of the objectives and so on... 

So in the end, we thought it would be better if the interfaced "optimizer" provides a list of what objectives are available and what configuration parameters they have, and SlicerRT just handles it on this level for display and returning the information to the optimizer. This also makes it very simple to interface a custom planning routine that works with some crazy objective prototypes.

Every idea is welcome here on how to have the most user- and developer-friendly situation here.

wahln:
@LiBu981 I remember the weird bug of SlicerRT crashing when the plan was created too quickly after starting the EBP Module. Is this still a thing?

LiBu981:
> @LiBu981 I remember the weird bug of SlicerRT crashing when the plan was created too quickly after starting the EBP Module. Is this still a thing?

Yes, it still occasionally crashes on creating a new plan.

ferdymercury:
> Every idea is welcome here on how to have the most user- and developer-friendly situation here.

Thanks Niklas for the feedback! I understand your point and motivation.

The drawback (from a user-friendly perspective) of delegating everything to the optimizer would be that, if I now want to change from EngineA or EngineB to test which one gives faster or better results, one has to redefine all objectives and constraints. This makes sense if EngineA does LET optimization and EngineB does dose rate optimization. However, for more 'common' examples, maybe the user just defined MinDose and MaxDose objectives, since they are not even aware of those other crazy objectives.

Now let's imagine that both EngineA and EngineB implement exactly the same number and type of objectives, but since they were developed by different persons, one calls these objectives `miniDose` and the other `MinDose`, and a third engine maybe `Min_Dose`. How should SlicerRT know that when I change from optimizer A to optimizer B, please remap those names to those other names in order to not remove the already set values in the GUI?

Maybe this is also a question of whether we want this to be optimization-research-oriented or newbie-user-oriented. Tradeoff and compromises.

Maybe there is a clever way of doing this, such as follows:
- Define consensus names / C++ enums? for common objectives+parameters with common optimizers
- Do what Niklas mentions of having everything delegated to the optimizer
- Force the optimizer to provide a mapping-file of 'their names' to the 'consensus enums' in case that that optimizer behaves in the typical way. That way, if I switch Engines, SlicerRT could see that if both EngineA and EngineB have minDose and Min_Dose objectives, but in their (respective) map JSON files, they associate that objective with the SlicerRT consensus enum, then it would copy-paste the settings to that one.

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -74,6 +74,14 @@ vtkMRMLRTBeamNode::vtkMRMLRTBeamNode()
   this->CollimatorAngle = 0.0;
   this->CouchAngle = 0.0;
 
+  this->DoseGridDim[0] = -1;
```
cpinter: Where are these (dimensions and spacing) set? I cannot find it. Also, dimensions and spacing are not enough to define a volume geometry, we also need orientation. I understand that for the Dij matrix it may not be needed (manybe neither spacing, not sure), but this seems to be hanging in the air here.

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -916,3 +924,138 @@ void vtkMRMLRTBeamNode::CreateMLCPointsFromSectionBorder(double jawBegin,
...
+    tripletList.push_back(T(rows[i], columns[i], values[i]));
+  }
+
+  this->DoseInfluenceMatrix = DoseInfluenceMatrixType(numRows, numCols);
+  this->DoseInfluenceMatrix.setFromTriplets(tripletList.begin(), tripletList.end());
+
+  // Store dose grid dimensions and spacing on which the dose influence matrix is defined
+  for (int i = 0; i < 3; ++i)
+  {
+	  if (doseGridDim == nullptr)
```
cpinter: We use 2 spaces for indentation in Slicer and SlicerRT, please fix it everywhere.
Also curly brackets start in a new line (here it is good, but in other places it is at the end of the previous line), please fix those too. Thanks!

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -916,3 +924,138 @@ void vtkMRMLRTBeamNode::CreateMLCPointsFromSectionBorder(double jawBegin,
...
+}
+
+//---------------------------------------------------------------------------
+int vtkMRMLRTBeamNode::GetDoseInfluenceMatrixRows()
+{
+  return this->DoseInfluenceMatrix.rows();
+}
+
+//---------------------------------------------------------------------------
+int vtkMRMLRTBeamNode::GetDoseInfluenceMatrixColumns()
```
cpinter: From the header I can see that this is for getting the number of rows, but the function name is a bit ambiguous. I'd use `GetDoseInfluenceMatrixColumnCount`, or `GetDoseInfluenceMatrixNumberOfColumns`. Same with the other one for rows.

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -916,3 +924,138 @@ void vtkMRMLRTBeamNode::CreateMLCPointsFromSectionBorder(double jawBegin,
...
+  indices->SetArray(this->DoseInfluenceMatrix.innerIndexPtr(), this->DoseInfluenceMatrix.nonZeros(), 1);
+  fieldData->AddArray(indices);
+
+  vtkSmartPointer<vtkIntArray> indptr = vtkSmartPointer<vtkIntArray>::New();
+  indptr->SetName("Indptr");
+  indptr->SetArray(this->DoseInfluenceMatrix.outerIndexPtr(), this->DoseInfluenceMatrix.outerSize() + 1, 1);
+  fieldData->AddArray(indptr);
+
+  return fieldData;
+}
```
cpinter: Please make sure every file has an empty line at the end. Otherwise we'll get warnings. Thanks

Beams/MRML/vtkMRMLRTBeamNode.h
```diff
@@ -248,7 +308,16 @@ class VTK_SLICER_BEAMS_MODULE_MRML_EXPORT vtkMRMLRTBeamNode : public vtkMRMLMode
   /// Couch angle
   double CouchAngle;
 
+  /// Dose influence matrix
+  DoseInfluenceMatrixType DoseInfluenceMatrix;
+
+  /// Dose grid dimensions (on which the dose influence matrix is defined)
+  double DoseGridDim[3];
```
cpinter: This should not be double. The get/set functions use int. I'd use unsigned int probably.

cpinter: Also now we can initialize values in the header like:
`unsigned int DoseGridDim {5, 5, 5};`

Beams/MRML/vtkMRMLRTPlanNode.cxx
```diff
@@ -57,6 +57,22 @@ vtkMRMLNodeNewMacro(vtkMRMLRTPlanNode);
...
+
+  this->TargetSegmentID = nullptr;
+
+  this->IsocenterSpecification = vtkMRMLRTPlanNode::CenterOfTarget;
+
+  this->NextBeamNumber = 1;
+
+  this->DoseEngineName = nullptr;
+
+  this->DoseGrid[0] = 5.0;
```
cpinter: Two things here
* If this is supposed to be dimensions, must indicate in variable name
* Seems to be a duplicate of `DoseGridDim` in the beam node. Use one or the other please

Beams/MRML/vtkMRMLRTPlanNode.cxx
```diff
@@ -57,6 +57,22 @@ vtkMRMLNodeNewMacro(vtkMRMLRTPlanNode);
...
+  this->DoseGrid[1] = 5.0;
+  this->DoseGrid[2] = 5.0;
+
+  this->IonPlanFlag = false;
```
cpinter: Remove this line. It is initialized in the header.

Beams/MRML/vtkMRMLRTPlanNode.cxx
```diff
@@ -217,6 +233,32 @@ void vtkMRMLRTPlanNode::SetDoseEngineName(const char* engineName)
...
+  else
+  {
+    this->PlanOptimizerName = nullptr;
+  }
+
+  // Invoke events
+  this->Modified();
+  this->InvokeEvent(vtkMRMLRTPlanNode::PlanOptimizerChanged, this);
+}
+
```
cpinter, line 252: Just one empty line is enough. Please ensure everywhere else too

Beams/MRML/vtkMRMLRTPlanNode.cxx
```diff
@@ -872,3 +914,34 @@ bool vtkMRMLRTPlanNode::ComputeTargetVolumeCenter(double center[3])
 
   return true;
 }
+
+//----------------------------------------------------------------------------
+void vtkMRMLRTPlanNode::setDoseGridInCoordinate(int index, double value)
```
cpinter: All function names must start with capital letter in C++ classes.

Beams/MRML/vtkMRMLRTPlanNode.cxx
```diff
@@ -872,3 +914,34 @@ bool vtkMRMLRTPlanNode::ComputeTargetVolumeCenter(double center[3])
...
+        vtkErrorMacro("setDoseGridInCoordinate: Invalid index");
+        return;
+    }
+
+    this->DoseGrid[index] = value;
+	this->Modified();
+}
+
+//----------------------------------------------------------------------------
+void vtkMRMLRTPlanNode::setDoseGridToCTGrid()
```
cpinter: "Grid" does not mean dimensions. Please rename the function to be less ambiguous.

Beams/MRML/vtkMRMLRTPlanNode.cxx
cpinter: Also the function name does not seem to be in sync with its purpose. You are setting one component of the dimensions of the grid, not a value in the grid.

Beams/MRML/vtkMRMLRTPlanNode.h
```diff
@@ -187,13 +190,26 @@ class VTK_SLICER_BEAMS_MODULE_MRML_EXPORT vtkMRMLRTPlanNode : public vtkMRMLFold
...
+  /// Set optimization engine name
+  void SetPlanOptimizerName(const char* optimizerName);
+
   /// Get prescription dose
   vtkGetMacro(RxDose, double);
   /// Set prescription dose
   vtkSetMacro(RxDose, double);
 
-  vtkSetVector3Macro(DoseGrid, double);
+  /// Get dose grid spacing
```
cpinter, line 209: I thought this was dimensions but now it says spacing. Getting really interesting :)

ExternalBeamPlanning/Logic/CMakeLists.txt
```diff
@@ -19,9 +19,14 @@ SET (${KIT}_INCLUDE_DIRS ${CMAKE_CURRENT_SOURCE_DIR} ${CMAKE_CURRENT_BINARY_DIR}
 set(${KIT}_TARGET_LIBRARIES
   vtkPlmCommon
   vtkSlicerRtCommon
+  vtkSlicerExternalBeamPlanningModuleMRML
   vtkSlicerMarkupsModuleMRML
   vtkSlicerBeamsModuleMRML
   vtkSlicerBeamsModuleLogic
+  vtkSlicer${MODULE_NAME}ModuleMRML
```
cpinter, line 25: You seem to be adding EBP MRML twice.

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.cxx
```diff
@@ -0,0 +1,103 @@
+/*==============================================================================
```
cpinter, line 1: Feel free to change the copyright too.

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.cxx
```diff
@@ -0,0 +1,103 @@
...
+
+
+//----------------------------------------------------------------------------
+vtkMRMLRTObjectiveNode::vtkMRMLRTObjectiveNode()
+{
+	this->Name = nullptr;
+}
+
+//----------------------------------------------------------------------------
+vtkMRMLRTObjectiveNode::~vtkMRMLRTObjectiveNode() {
```
cpinter: Code formatting is quite different from the conventions in this class. Please fix coding style everywhere according to previous comments (and Slicer style)

ExternalBeamPlanning/CMakeLists.txt
```diff
@@ -20,11 +21,18 @@ set(MODULE_EXPORT_DIRECTIVE "Q_SLICER_QTMODULES_${MODULE_NAME_UPPER}_EXPORT")
 
 set(MODULE_INCLUDE_DIRECTORIES
   ${SlicerRtCommon_INCLUDE_DIRS}
+  ${CMAKE_CURRENT_SOURCE_DIR}/MRML
```
cpinter: Why are these necessary?

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.cxx
```diff
@@ -0,0 +1,103 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+This file was originally developed by Lina Bucher, Institut fuer Biomedizinische
+Technik at the Karlsruher Institut fuer Technologie (IBT-KIT) and German Cancer
+Research Center (DKFZ)
+
+==============================================================================*/
+
+// MRMLObjective includes
```
cpinter: There is no project called `MRMLObjective`. This is `ExternalBeamPlanning MRML`.

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
```diff
@@ -0,0 +1,72 @@
...
+
+  /// Get unique node XML tag name (like Volume, Model) 
+  const char* GetNodeTagName() override { return "Objective"; };
+
+public:
+  vtkGetStringMacro(Name);
+  vtkSetStringMacro(Name);
+
+  /// Set name of segmentation assigned to this objective
+  void SetSegmentation(std::string segmentationName);
```
cpinter: A few major problems with this:
* You are referencing to another node by name if I understand correctly? It is very error-prone because nodes can be renamed super easily. Even IDs cannot be used because they change on scene import. Please use node references like in other classes.
* Segmentation is supposed to mean the `vtkSegmentation` class. But you call it the name of a node. Please do not mix things this much

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
```diff
@@ -0,0 +1,72 @@
...
+  /// Get name of segmentation assigned to this objective
+  std::string GetSegmentation();
+
+protected:
+  vtkMRMLRTObjectiveNode();
+  ~vtkMRMLRTObjectiveNode() override;
+  vtkMRMLRTObjectiveNode(const vtkMRMLRTObjectiveNode&);
+  void operator=(const vtkMRMLRTObjectiveNode&);
+
+  char* Name;
```
cpinter: Just use std::string, much easier.

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
```diff
@@ -0,0 +1,72 @@
...
+  /// Name of the segmentation assigned to this objective
+  std::string Segmentation;
```
cpinter: Node reference, very important!

ExternalBeamPlanning/Widgets/CMakeLists.txt
```diff
@@ -42,8 +68,22 @@ set(${KIT}_MOC_SRCS
   qSlicerDoseEngineLogic.h
   qSlicerMockDoseEngine.h
   qSlicerScriptedDoseEngine.h
+  qSlicerAbstractPlanOptimizer.h
+  qSlicerPlanOptimizerPluginHandler.h
+  qSlicerPlanOptimizerLogic.h
```
cpinter, line 74: It is strange to have classes called Logic in the Widgets project. I'm sure there is a good reason, let's discuss this later.

ExternalBeamPlanning/Widgets/CMakeLists.txt
```diff
@@ -85,4 +128,5 @@ SlicerMacroBuildModuleWidgets(
 #-----------------------------------------------------------------------------
 if(Slicer_USE_PYTHONQT)
   add_subdirectory(Python)
+  #add_subdirectory(Plugins)
```
cpinter: Delete unused line.

ExternalBeamPlanning/Widgets/Python/CMakeLists.txt
```diff
@@ -1,25 +1,14 @@
...
-  AbstractScriptedDoseEngine
-  MockPythonDoseEngine
-  EGSnrcUtil
-  OrthovoltageDoseEngineUtil
-  OrthovoltageDoseEngine
-  )
-
-set(DoseEngines_PYTHON_RESOURCES
+set(Python_PYTHON_SCRIPTS
+  prepareRTDataset.py
```
cpinter: Rename this file, see [comment](https://github.com/SlicerRt/SlicerRT/pull/275#issuecomment-3151288716).

cpinter (changes requested):
I did a first round of review. Please fix the obvious ones and let's discuss anytihng that is not obvious. Thank you very much!

Beams/MRML/vtkMRMLRTBeamNode.cxx
```diff
@@ -916,3 +924,138 @@ void vtkMRMLRTBeamNode::CreateMLCPointsFromSectionBorder(double jawBegin,
...
+  // Store dose grid dimensions and spacing on which the dose influence matrix is defined
+  for (int i = 0; i < 3; ++i)
+  {
+	  if (doseGridDim == nullptr)
+	  {
+		  this->DoseGridDim[i] = -1;
+	  }
+      else
+	  {
+		  this->DoseGridDim[i] = doseGridDim[i];
```
LiBu981, line 958: Concerning [the setting of DoseGridDim (and DoseGridSpacing) in the beamNode](https://github.com/SlicerRt/SlicerRT/pull/275/files#r2260505403):

Currently it is saved here, if passed as input when a dose influence matrix is saved in the beamNode.

This is used for example in the [pyRadPlanEngine.py](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-5452cf4fe3b756152c3c6419276c9c50c09b0acb08f3701abd6a38f5447ccc70R164-R170) whereby the _dij.dose_grid_ is defined in the [prepareRTDataset.py](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-9c9841bb44ae2060d35d7b5c2a84210fa6f461e7acedd0b6759948917f3d38c9R56-R73).

I added this to be able to resample the dose in the [MockOptimizer](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-495b22737b8113593170b35b81f9964e7cdf9ec633660092162af7f74e28fb98R111-R190). Maybe this can be handled differently though?

Beams/MRML/vtkMRMLRTBeamNode.cxx
LiBu981: Currently it is saved [here ](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-f01f42da398904b09d5f94ac5a17b7829cf8c80ff12854d98e67d8e7afed277cR949-R971), if passed as input when a dose influence matrix is saved in the beamNode.

This is used for example in the [pyRadPlanEngine.py](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-5452cf4fe3b756152c3c6419276c9c50c09b0acb08f3701abd6a38f5447ccc70R164-R170) whereby the _dij.dose_grid_ is defined in the [prepareRTDataset.py](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-9c9841bb44ae2060d35d7b5c2a84210fa6f461e7acedd0b6759948917f3d38c9R56-R73).

I added this to be able to resample the dose in the [MockOptimizer](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-495b22737b8113593170b35b81f9964e7cdf9ec633660092162af7f74e28fb98R111-R190). Maybe this can be handled differently though?

Beams/MRML/vtkMRMLRTBeamNode.cxx
LiBu981: Should be fixed now :)
(see [commit a0c435d ](https://github.com/SlicerRt/SlicerRT/pull/275/commits/a0c435d30031a36cd1d650f065909536993d1a78))

Beams/MRML/vtkMRMLRTBeamNode.cxx
LiBu981: Should be fixed now :)
(see [commit a0c435d ](https://github.com/SlicerRt/SlicerRT/pull/275/commits/a0c435d30031a36cd1d650f065909536993d1a78))

ExternalBeamPlanning/Widgets/CMakeLists.txt
LiBu981: Deleted :)
(see [commit 2887ded](https://github.com/SlicerRt/SlicerRT/pull/275/commits/2887dedb111ab1ae04b670135bc0cac07bbef5b3))

ExternalBeamPlanning/Widgets/Python/CMakeLists.txt
LiBu981: Done 👍 ([commit 60e59e4](https://github.com/SlicerRt/SlicerRT/pull/275/commits/60e59e4ee94cfb74356e55f09f627ddf90b7f118))

Beams/MRML/vtkMRMLRTBeamNode.cxx
LiBu981: Done 👍  ([commit 7d2208f](https://github.com/SlicerRt/SlicerRT/pull/275/commits/7d2208f6aa3223e1fe882b5421c751d354968b99))

Beams/MRML/vtkMRMLRTBeamNode.h
LiBu981: I changed it to **int** now, as I initialize here with {-1, -1, -1} to check weather a _DoseGridDim_ (same with _DoseGridSpacing_) was set. So **unsigned int** wouldn't work with this logic.
(e.g. in [pyRadPlanUtils](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-c82da6a5e12831f131c33b7b9aecdf2f7377b9a9cc0b5720999d19717df4037fR65-R75) when called from the [pyRadPlan optimizer](https://github.com/SlicerRt/SlicerRT/pull/275/files#diff-550f63a7090f885b0830c23c6a9edef06fb451410975c87d7a4bd163ca07d207R100-R101) )

([commit 5c0d8e2c](https://github.com/SlicerRt/SlicerRT/pull/275/commits/5c0d8e2c58d33a4f547c568014100e8578e3a8fb))

Beams/MRML/vtkMRMLRTPlanNode.cxx
LiBu981: It is supposed to be the grid spacing. I changed it to _DoseGridSpacing_ ([commit 5c0d8e2c](https://github.com/SlicerRt/SlicerRT/pull/275/commits/5c0d8e2c58d33a4f547c568014100e8578e3a8fb))

Beams/MRML/vtkMRMLRTPlanNode.cxx
LiBu981: Renamed it to _SetDoseGridSpacingComponent_ ([commit 5c0d8e2c](https://github.com/SlicerRt/SlicerRT/pull/275/commits/5c0d8e2c58d33a4f547c568014100e8578e3a8fb)). Does this work?

Beams/MRML/vtkMRMLRTPlanNode.cxx
LiBu981: Renamed to _SetDoseGridSpacingToCTGridSpacing_ ([commit 5c0d8e2c](https://github.com/SlicerRt/SlicerRT/pull/275/commits/5c0d8e2c58d33a4f547c568014100e8578e3a8fb)).

Beams/MRML/vtkMRMLRTPlanNode.h
LiBu981: Yes, sorry, that was a little confusing. Should be clearer now ;) ([commit 5c0d8e2c](https://github.com/SlicerRt/SlicerRT/pull/275/commits/5c0d8e2c58d33a4f547c568014100e8578e3a8fb)).

Beams/MRML/vtkMRMLRTPlanNode.cxx
LiBu981: Removed 👍 ([commit 8371b50](https://github.com/SlicerRt/SlicerRT/pull/275/commits/8371b50c9be82d5f1431352b75f57b485f64deed))

Beams/MRML/vtkMRMLRTPlanNode.cxx
LiBu981: But yes, _doseGridSpacing_ is both in the _planNode_ and the _beamNode_. I added it in the _beamNode_ to be able to keep track of the dose grid used to calculate the dose influence matrix saved in the _beamNode_.

This was necessary for example in the **MockOptimizer** to resample the dose to the _referenceVolume_.
In the **pyRadPlanUtils** it is also used to ensure that the optimization uses the same dose grid as in the dose calculation to prevent any errors when the dose grid is changed in the UI inbetween.

Maybe this can be handled better though?

Beams/MRML/vtkMRMLRTPlanNode.cxx
LiBu981: I deleted it here, but will keep my eyes open for other cases 👍 ([commit 8371b50](https://github.com/SlicerRt/SlicerRT/pull/275/commits/8371b50c9be82d5f1431352b75f57b485f64deed)).

ExternalBeamPlanning/CMakeLists.txt
LiBu981: Deleted ([commit c6dd3a5](https://github.com/SlicerRt/SlicerRT/pull/275/commits/c6dd3a55503ec129adab2345a39b556ee1529ec6))

ExternalBeamPlanning/Logic/CMakeLists.txt
LiBu981: Removed ([commit 62ee16b](https://github.com/SlicerRt/SlicerRT/pull/275/commits/62ee16b396f02ec91aa51e9f7dcd44c5d505eca3))

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.cxx
LiBu981: I did some renaming of the include comments ([commit 3bb6baa](https://github.com/SlicerRt/SlicerRT/pull/275/commits/3bb6baadcea5d225b7d03669a03f4bc110b413da)). Does this fit the conventions?

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
LiBu981: Okay, I tried fixing this by replacing the segmentation _name reference_ in the objective node to a reference to the _segmentationNode_ and the _segmentID_ and then, when setting these in the objective Node , check whether the ID is present in the node.
Is this a good approach?
([commit be3969d](https://github.com/SlicerRt/SlicerRT/pull/275/commits/be3969d8773451dfe406c414a6a10f89d1f10ddf))

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
LiBu981: I used **char*** so that _vtkGetStringMacro(Name)_ works? Is there a way to do this with **std::string**?
(I oriented myself on the beam node.)

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
LiBu981: See [comment](https://github.com/SlicerRt/SlicerRT/pull/275#discussion_r2285291056) :)

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
```diff
@@ -0,0 +1,680 @@
...
+//-----------------------------------------------------------------------------
+void qMRMLObjectivesTableWidget::setPlanNode(vtkMRMLNode* node)
+{
+  Q_D(qMRMLObjectivesTableWidget);
+
+  vtkMRMLRTPlanNode* planNode = vtkMRMLRTPlanNode::SafeDownCast(node);
+
+  // Connect plan modified events to population of the table
+  qvtkReconnect(d->PlanNode, planNode, vtkCommand::ModifiedEvent, this, SLOT(updateObjectivesTable()));
+
```
LiBu981: I'm struggling a bit to add a connection between the _SegmentationNode_ in the **Plan** and in the **Objectives Table**, so that on modifications of the selected segmentation (e.g. name of a segment) the selectable segments in the objectives table are modified as well. (I tried looking into the _qMRMLBeamsTableView.cxx_ for inspiration.)
Does someone have ideas?

MichaelColonel: If you don't know the type of modification (for example segment has been added, deleted or renamed) then just clean a table and refill it with new information for each segment.

MichaelColonel: If position of the segment ID with segmentation node is unchanged, then just update corresponding table items according to the position of the segment within table.

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.cxx
LiBu981: See [commit c803da7](https://github.com/SlicerRt/SlicerRT/pull/275/commits/c803da79f8827e3697288ba3b54b688dd9921036)

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
```diff
@@ -0,0 +1,680 @@
+/*==============================================================================
+
+Copyright (c) Laboratory for Percutaneous Surgery (PerkLab)
```
cpinter: Feel free to change the copyright in new files to your institution

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
```diff
@@ -0,0 +1,680 @@
...
+    qCritical() << Q_FUNC_INFO << ": Invalid segmentation node";
+    return;
+  }
+  vtkSegmentation* segmentation = segmentationNode->GetSegmentation();
+  std::vector<std::string> segmentIDs;
+  segmentationNode->GetSegmentation()->GetSegmentIDs(segmentIDs);
+
+  for (const std::string& segmentID : segmentIDs)  
+  {  
+      vtkSegment* segment = segmentation->GetSegment(segmentID);
```
cpinter: Indentation is two spaces please

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
```diff
@@ -0,0 +1,680 @@
...
+  QComboBox* objectivesDropdown = qobject_cast<QComboBox*>(d->ObjectivesTable->cellWidget(row, d->columnIndex("ObjectiveName")));
+  QComboBox* segmentationsDropdown = qobject_cast<QComboBox*>(d->ObjectivesTable->cellWidget(row, d->columnIndex("Segments")));
+
+  QString segmentID = segmentationsDropdown->currentData().toString();
+  if (segmentID.isEmpty())
+  {
+    qCritical() << Q_FUNC_INFO << ": Segment ID is empty!";
+    return;
+  }
+		vtkMRMLSegmentationNode* segmentationNode = d->PlanNode->GetSegmentationNode();
```
cpinter: Fix formatting here

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
cpinter: In general I agree with @MichaelColonel in that in case of such custom table widgets, the simplest is to re-populate the table on relevant changes. The table will never contain more than a few dozen rows, so speed is not an issue. In general there are two ways to implement tables that sync with the MRML scene:
1. Qt model-view pattern: see `qMRMLSubjectHierarchyModel` and `qMRMLSubjectHierarchyTreeView` for example. It seems complicated, but it is the correct way
2. However, sometimes a simple table widget is enough, especially if there is not a lot of data to handle (unlike the whole MRML scene can be). See functions starting with `populate` for example. I think even the SegmentsTableView worked like that in the beginning.

Your solution seems to be halfway between the two, in that you handle addition, deletion, etc. So either simplify it to always re-populate (but it's work done twice, and more time), or do what @MichaelColonel suggests second, i.e. observe the segment modified event of the segmentation node, then look for its row in the table, and update the item text. I hope this helped a bit. If not, happy to meet Friday morning to discuss (this and other questions too).

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
LiBu981: See [commit c803da7](https://github.com/SlicerRt/SlicerRT/pull/275/commits/c803da79f8827e3697288ba3b54b688dd9921036)

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
LiBu981: See [commit e60fd9f](https://github.com/SlicerRt/SlicerRT/pull/275/commits/e60fd9ffd2fd7e561c49a8f93ca7b15585d90f28)

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
LiBu981: Sorry, I thought I had fixed this already, but some places it were still tabs instead of spaces. Should be fixed now.
(See [commit 5b22bd3](https://github.com/SlicerRt/SlicerRT/pull/275/commits/5b22bd3d96e63161e4f66c63c1869f5a15aa5c81))

ExternalBeamPlanning/MRML/vtkMRMLRTObjectiveNode.h
LiBu981: I have changed the saving of the Segmentation to a node reference now.
(See [commit 2f61b50](https://github.com/SlicerRt/SlicerRT/pull/275/commits/2f61b504efa15b765d496dd0155df303a7f2867f))

cpinter:
Hi @LiBu981 I tried to build this branch but I got build errors. When I tried to rebase it to the latest SlicerRT I immediately got conflicts, and since you have dozens of commits and I'm not that familiar with the content and the reason of the choices you made, I'd like to kindly ask you to do the rebase on the latest SlicerRT so that the branch is up-to-date. Thank you!

LiBu981:
Is rebased now :)

LiBu981:
This should fix the crash: [commit 2dc6206](https://github.com/SlicerRt/SlicerRT/pull/275/commits/2dc6206346639eec66f0c5e76b54944917238cdc)
The issue was that the _PlanOptimizerName_ wasn't initialized in the _PlanNode_. When setting the _DoseEngineName_ the _updatePlanOptimizer_ was triggered (over a _ModifiedEvent_) and tried to call the optimizer by name.

I haven't had any crashes now.

cpinter:
Thank you so much @LiBu981 ! I'll build it today and start reviewing (and understanding) it in more detail

cpinter:
Just a quick comment. I think you may have done the rebase in the reverse order, because the upstream commits and your commits are intertwined.

It will be quite hard to integrate it this way, and also I cannot squash your commits to see the changes overall.
I can build and test it, but unfortunately I cannot really review it this way. Can you please do the rebase in the habitual way, i.e. all your commits are on top of all the upstream commits?

image

Thank you very much!

LiBu981:
Hi everyone, sorry for the delay.

It should be properly rebased this time! I was in a bit of a rush last week and apparently didn't think about what I was doing. Sorry for the confusion.

Hope this is now all correct. I used @ferdymercury 's rebase (thank you for the help!!) and added some corrections. It is running through for me now.

cpinter:
Thank you so much @LiBu981 ! I'm also reacting a bit slow; yesterday we had a very important whole-day visit and had to prepare etc... I'm doing a thorough review now and trying how it works. I'll push some trivial fixes and TODOs in subsequent commits to this branch.

cpinter:
Update: This branch does not build currently as is. I think some merge conflicts have been left unfixed (at least I found a harmless one already). I'll try to push the fix ASAP.

cpinter:
I thought I could push to branches that are in PR and I'm added as a reviewer, but apparently not. I pushed one commit that fixes the build here: https://github.com/cpinter/SlicerRT/tree/pyRadPlanIntegration_dev/dev

ferdymercury:
> I thought I could push to branches that are in PR and I'm added as a reviewer, but apparently not. I pushed one commit that fixes the build here: https://github.com/cpinter/SlicerRT/tree/pyRadPlanIntegration_dev/dev

For that to work, @LiBu981 should check this box on the right of the PR panel:

image

LiBu981:
> > I thought I could push to branches that are in PR and I'm added as a reviewer, but apparently not. I pushed one commit that fixes the build here: https://github.com/cpinter/SlicerRT/tree/pyRadPlanIntegration_dev/dev
> 
> For that to work, @LiBu981 should check this box on the right of the PR panel:
> 
> image


The box was already checked :)
{B30AFB49-637D-45DB-A094-C43C5B9258A9}

I think the problem is that the branch is a fork. I've added you as contributors to my repository, can you commit now?
Otherwise, I can include the commits.

ferdymercury:
> I think the problem is that the branch is a fork. I've added you as contributors to my repository, can you commit now?
> Otherwise, I can include the commits.

Hmmm the problem might actually be that your repository is a fork of https://github.com/e0404/SlicerRT rather than a fork of https://github.com/SlicerRt/SlicerRT

I think the easiest would be to create a local backup, erase your repo from GH, and fork again cleanly from the SlicerRt, and then push back from local copy your branches to upstream. But I think there are other ways such as https://github.com/cli/cli/discussions/10956

cpinter:
Thanks a lot @LiBu981 I was able to push this commit now!

Strange, because when we do a PR in Slicer, the people with push rights (JC, Andras, ...) can push directly to the topic branches (which are from forks). Anyway, good for now :) Thanks again

ferdymercury:
> Strange, because when we do a PR in Slicer, the people with push rights (JC, Andras, ...) can push directly to the topic branches (which are from forks)

yep. Maybe due to of the fork-of-fork structure (see above).

cpinter:
Ah OK I didn't catch that. Thanks @ferdymercury

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.h
```diff
@@ -0,0 +1,147 @@
...
+  virtual QString name()const;
+  /// Set the name of the Optimization engine
+  /// NOTE: name must be defined in constructor in C++ engines, this can only be used in python scripted ones
+  virtual void setName(QString name);
+
+public:
+  /// Structure to represent objective functions (only name and parameters)
+  struct ObjectiveStruct
+  {
+    std::string name;
```
cpinter: As a Qt class, it would make sense to use QString and QMap. Is there any reason not to use those?

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.h
```diff
@@ -0,0 +1,147 @@
...
+    std::map<std::string, std::string> parameters;
+  };
+
+signals:
+  void progressInfoUpdated(QString info);
+
+// Optimization calculation related functions
+public:
+  /// Perform Optimization calculation for a single beam
+  /// \param Beam node for which the Optimization is calculated
```
cpinter: This line seems to be wrong. It says beam node but the argument is plan node.

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.h
```diff
@@ -0,0 +1,147 @@
...
+    std::string name;
+    std::map<std::string, std::string> parameters;
+  };
+
+signals:
+  void progressInfoUpdated(QString info);
+
+// Optimization calculation related functions
+public:
+  /// Perform Optimization calculation for a single beam
```
cpinter: "for a single beam" => seems otherwise (i.e. whole plan)

ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
```diff
@@ -0,0 +1,175 @@
+from DoseEngines import *
+from Python.PyRadPlanUtils import prepareCt, prepareCst, preparePln
+
+class pyRadPlanEngine(AbstractScriptedDoseEngine):
+  """ Mock python dose engine to test python interface for External Beam Planning
```
cpinter: Fix description please

ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
```diff
@@ -0,0 +1,175 @@
...
+    
+    self.scriptedEngine.updateBeamParameterComboBox(
+    "pyRadPlan parameters", "radiationMode", parameter_label,
+    "comment", available_radiation_modes, 0)
+
+
+  def calculateDoseUsingEngine(self, beamNode, resultDoseVolumeNode):
+
+    #################################### Import pyRadPlan libraries ######################################
+    import logging
```
cpinter: As far as I know the good practice is to import everything on the top of the file

cpinter:
Finishing for today. Sorry for the many comments here. I also added some in the code, they are minor fixes.

One thing that I'd like to ask in general is to review the usage of smart pointers. In some cases it is not needed, for example in `qSlicerMockPlanOptimizer::optimizePlanUsingOptimizer` (I'd suggest reviewing their purpose: https://slicer.readthedocs.io/en/5.4/developer_guide/advanced_topics.html#memory-management , https://www.kitware.com/a-tour-of-vtk-pointer-classes/). In general, this is the main point:

> A vtkSmartPointer pointer can be created without actually creating an object, so vtkSmartPointer should be used when:
> * we don’t know the exact object type that we want to create at the time of the pointer creation (e.g., we create a vtkSmartPointer<vtkMRMLVolumeNode> and later set it to point to a vtkMRMLScalarVolumeNode or vtkMRMLVectorVolumeNode), or
> * we need to take the ownership of an already created object (using vtkSmartPointer::Take(...)

Thanks a lot again! I'll finish reviewing the code probably tomorrow and will finally try it out in action. I think we are close to integration!

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.h
LiBu981: Thanks for spotting! It was left over from the Dose Engine. I hope I've caught any wrong descriptions now (see [commit d7e8297](https://github.com/SlicerRt/SlicerRT/pull/275/commits/d7e829718276346566aee0540330e7547297a1c3)).

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.h
LiBu981: As above :) (see [commit d7e8297](https://github.com/SlicerRt/SlicerRT/pull/275/commits/d7e829718276346566aee0540330e7547297a1c3))

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.h
LiBu981: I thought I had problems before with some conversions here, but I've changed it to Qt types now without any issues.  So not entirely sure why I didn't use it before.... (see [commit 7acb67b](https://github.com/SlicerRt/SlicerRT/pull/275/commits/7acb67bf035fa45ab8978fd28bebf863894f1f00))

ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
LiBu981: Yes, I think I can put everything except the pyRadPlan internal imports outside/ at the top of the file. Otherwise, we will get an error on opening Slicer before pyRadPlan is installed the first time (with the popup install request).

LiBu981: See [commit 1b4bfcf](https://github.com/SlicerRt/SlicerRT/pull/275/commits/1b4bfcf891fa617075d59519f9abb297304c76aa)

ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
LiBu981: See [commit 1b4bfcf](https://github.com/SlicerRt/SlicerRT/pull/275/commits/1b4bfcf891fa617075d59519f9abb297304c76aa)

LiBu981:
I've reviewed the use of the vtkSmartPointer and hope I've caught any wrong uses (see [commit b2a87](https://github.com/SlicerRt/SlicerRT/pull/275/commits/b2a87e4d0be2aff0bd2045d224cfa707c9b10e26)).

Though, if I saw correctly, we normally use a vtkSmartPointer for the vtkImageData before passing it to SetAndObserveImageData() (e.g. in `qSlicerMockDoseEngine::calculateDoseUsingEngine`), so I kept that in `qSlicerMockPlanOptimizer::optimizePlanUsingOptimizer` for the _totalDoseImageData_.

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.cxx
```diff
@@ -146,15 +146,13 @@ QString qSlicerAbstractPlanOptimizer::optimizePlan(vtkMRMLRTPlanNode* planNode)
...
-  // (Smart pointers are used to auto-delete references so that memory is freed, but it is not managed here)
-  vtkSmartPointer<vtkMRMLScalarVolumeNode> resultOptimizationVolumeNode = planNode->GetOutputTotalDoseVolumeNode();
+  vtkMRMLScalarVolumeNode* resultOptimizationVolumeNode = planNode->GetOutputTotalDoseVolumeNode();
   if (!resultOptimizationVolumeNode)
   {
-   resultOptimizationVolumeNode = vtkSmartPointer<vtkMRMLScalarVolumeNode>::New();
-      // Give default name for result node (engine can give it a more meaningful name)
-      std::string resultOptimizationNodeName = std::string(planNode->GetName()) + "_Optimization";
-      resultOptimizationVolumeNode->SetName(resultOptimizationNodeName.c_str());
+    resultOptimizationVolumeNode = vtkMRMLScalarVolumeNode::New();
```
cpinter: This will result in a memory leak. Please use `vtkNew`, otherwise this reference will be dangling when the scene is cleared and the object will not be deleted from memory.

ExternalBeamPlanning/Widgets/qSlicerMockPlanOptimizer.cxx
```diff
@@ -64,7 +64,7 @@ QString qSlicerMockPlanOptimizer::optimizePlanUsingOptimizer(vtkMRMLRTPlanNode*
   }  
 
   // Create total dose image data from reference volume
-  vtkImageData* totalDoseImageData = vtkImageData::New();
+  vtkSmartPointer<vtkImageData> totalDoseImageData = vtkSmartPointer<vtkImageData>::New();
```
cpinter: Better to use `vtkNew` here as well.

cpinter:
I just pushed a really minor commit. I hope I'll have more time for this in the next week - I haven't been lucky lately... Thanks for your patience.

Let me know if there is anything in particular I should look at. Thank you!

ExternalBeamPlanning/Widgets/qMRMLObjectivesTableWidget.cxx
LiBu981: Okay, thank you so much for all the feedback. I finally managed to get to this and implemented reactions to changes in the _segmentationNode_ and the _planNode_ and it seems to work properly, though I would appreciate a review (see [commit 4215189](https://github.com/SlicerRt/SlicerRT/pull/275/commits/421518931453e89604b663235219818434c97698)). I hope this corresponds to your comments :) Thanks in advance!

LiBu981:
Thank you a lot @cpinter! I think a review on the objective table's reactions to changes in the _plan_ and _segmentation node_ [see here](https://github.com/SlicerRt/SlicerRT/pull/275#discussion_r2429379771) would be good, as I'm not 100% sure if I implemented everything correctly.

There are a few minor issues I wanted to fix the coming days as well as integrate your comments, though from my side this was the biggest ToDo still open.

ExternalBeamPlanning/Widgets/qSlicerAbstractPlanOptimizer.cxx
LiBu981: Okay I changed it :) (see [commit 3acf2c2](https://github.com/SlicerRt/SlicerRT/pull/275/commits/3acf2c2f19cbbc57662e26b79d90a25c6ce9f4fe))

ExternalBeamPlanning/Widgets/qSlicerMockPlanOptimizer.cxx
LiBu981: see [commit 3acf2c2](https://github.com/SlicerRt/SlicerRT/pull/275/commits/3acf2c2f19cbbc57662e26b79d90a25c6ce9f4fe)

cpinter:
Thanks a lot for keeping working on this! I just tried the branch and build fails due to some API change in CTK (I already fixed it but havent pushed), but then there is a crash at startup with this call stack:

```
>	qSlicerExternalBeamPlanningModuleWidgets.dll!qSlicerScriptedDoseEngine::setPythonSource(const QString newPythonSource) Line 150	C++
 	qSlicerExternalBeamPlanningModuleWidgets.dll!qSlicerScriptedDoseEngine::qt_static_metacall(QObject * _o, QMetaObject::Call _c, int _id, void * * _a) Line 84	C++
 	qSlicerExternalBeamPlanningModuleWidgets.dll!qSlicerScriptedDoseEngine::qt_metacall(QMetaObject::Call _c, int _id, void * * _a) Line 123	C++
 	PythonQt.dll!PythonQtSlotInfo::invokeQtMethod(QObject * obj, PythonQtSlotInfo * slot, void * * args) Line 669	C++
 	PythonQt.dll!PythonQtCallSlot(PythonQtClassInfo * classInfo, QObject * objectToCall, _object * args, bool strict, PythonQtSlotInfo * info, void * firstArgument, _object * * pythonReturnValue, void * * directReturnValuePointer, PythonQtPassThisOwnershipType * passThisOwnershipToCPP) Line 172	C++
 	PythonQt.dll!PythonQtSlotFunction_CallImpl(PythonQtClassInfo * classInfo, QObject * objectToCall, PythonQtSlotInfo * info, _object * args, _object * kw, void * firstArg, void * * directReturnValuePointer, PythonQtPassThisOwnershipType * passThisOwnershipToCPP) Line 446	C++
 	PythonQt.dll!PythonQtMemberFunction_Call(PythonQtSlotInfo * info, _object * m_self, _object * args, _object * kw) Line 277	C++
 	PythonQt.dll!PythonQtSlotFunction_Call(_object * func, _object * args, _object * kw) Line 265	C++
 	[External Code]	
 	PythonQt.dll!PythonQt::evalScript(_object * object, const QString & script, int start) Line 1019	C++
 	PythonQt.dll!PythonQtObjectPtr::evalScript(const QString & script, int start) Line 56	C++
 	qSlicerExternalBeamPlanningModule.dll!qSlicerExternalBeamPlanningModule::setup() Line 148	C++
 	qSlicerBaseQTCore.dll!qSlicerAbstractCoreModule::initialize(vtkSlicerApplicationLogic * _appLogic) Line 103	C++
 	qSlicerBaseQTCore.dll!qSlicerModuleFactoryManager::loadModule(const QString & name, const QString & dependee) Line 184	C++
 	qSlicerBaseQTCore.dll!qSlicerModuleFactoryManager::loadModule(const QString & name) Line 116	C++
 	SlicerApp-real.exe!qSlicerApplicationHelper::postInitializeApplication<qSlicerAppMainWindow>(qSlicerApplication & app, QScopedPointer<QSplashScreen,QScopedPointerDeleter<QSplashScreen>> & splashScreen, QScopedPointer<qSlicerAppMainWindow,QScopedPointerDeleter<qSlicerAppMainWindow>> & window) Line 299	C++
 	SlicerApp-real.exe!`anonymous namespace'::SlicerAppMain(int argc, char * * argv) Line 52	C++
 	SlicerApp-real.exe!main(int argc, char * * argv) Line 55	C++
```

I'm investigating it and if I can fix it I will push it, but the crash is at setting the python source of the new engine (the `newPythonSource` variable in the innermost call has the value `C:/d/_Extensions/SlicerRT_D/inner-build/lib/Slicer-5.9/qt-scripted-modules/DoseEngines/pyRadPlanEngine.py`), so we will see where the fix is needed.

Just wanted to give an update.

cpinter:
FYI it seems that the crash is related to the fact that the dose engine imports packages at a time when the Python environment is not completely initialized. We'll reorganize the instantiation of the scripted engines a bit so that it is done right after startup. I'll make a commit hopefully in the next few days about this.

cpinter:
Also FYI I just pushed the fix for the build error with the latest Slicer. Please let me know if it breaks anything on your side.

LiBu981:
It runs through for me. Should I rebase?

LiBu981:
One more question concerning the repopulating of the objectives table:

Currently when the plan is changed, the table is populated with the objectives saved in the optimizer (see [commit 508d282](https://github.com/SlicerRt/SlicerRT/pull/275/commits/508d2826ecf2d11fd38ee2c35c6393df6861f242#diff-e11baaf2df0641afe90ce5ba690f1edfb9cb431f9ef24aa5d641ad10eb9d86bcR245-R342)).
But this means, when plans use the same optimizer, they share the same objectives table.

How could this be handled differently? Should we save the objectives in the plan?

image
image
image<br/>
<br/>
image
image<br/>
<br/>
image
(The objectives table from RTPlan_1 overwrote the one in RTPlan)

ferdymercury:
> Should we save the objectives in the plan?

Thanks.
From my point of view, I guess this is better, since the DICOM RTplan establishes / allows that the objectives are specific of each RTplan https://dicom.innolitics.com/ciods/rt-ion-plan/rt-prescription/300a0010/300a0025

If there is a benefit for it, one could think of a button to "transfer objectives" from one RT plan to another?

This goes a bit in the direction that I advocate, that the most simple objectives should be generic (C++ enum or consensus string, or dictionary mapping how is eg MinDose-A spelling in B) so that they can be shared between optimizers, ie changing from optimizer A to optimizer B for the same RTplan should adopt (most) of the defined objectives.

cpinter:
I met earlier today with @LiBu981, who showed me the current funcionalities of the PR (I still haven't been able to try it yet due to the build error and then the startup crash, and did not want to integrate without seeing how it worked).

The branch is now integrated in the single commit 4c058cd82befb7b949a3826b38970e5413e97c01 (with @LiBu981 as main author and co-authors @wahln and @ferdymercury).

I also added the commits for the build error, and the startup crash. Unfortunately the crash is still there, but at least now instead of startup it happens when entering the EBP module. Most probably it is due to the updated PythonQt. I added a new issue for that: https://github.com/SlicerRt/SlicerRT/issues/284
I'll add more details to it in a bit, and we'll track the crash there.

I'm not yet closing the PR because there is a discussion about objective ownership. Note that the PR now mentions merge conflicts, which can be disregarded. @LiBu981 please do not push more commits to this branch, but make a new one from `master` and push to there. Feel free to close the PR whenever ready.

Thank you so much @LiBu981 and @wahln for this great contribution!

cpinter:
Closing because this branch has been integrated. For reference, a follow-up PR: https://github.com/SlicerRt/SlicerRT/pull/287

PR #276, opened 2025-07-21
FIX: avoid compilation warnings with latest Slicer and VTK9.5

ferdymercury:


MichaelColonel:
I've tested and it compiles successfully, but SlicerRT crushes on loading screen. Latest Slicer runs OK, but without SlicerRT.

ferdymercury:
Maybe related https://discourse.slicer.org/t/slicerrt-built-from-source-cannot-start/43792/12

ferdymercury:
Solved by @cpinter in master, thanks!

PR #280, opened 2025-09-11
BUG: Removed singleton tag for DRR markups

MichaelColonel:
That fixes a bug when markups don't show up after the scene has been closed, and a new data loaded.

Mouse interaction with DRR markups is disabled also.


cpinter:
Looks great, thank you!

Again, sorry for the delay. Feel free to ping me if things don't get attention - I have too many projects...

PR #283, opened 2025-10-30
FIX: prevent nullptr access if GetSegmentCenterRAS returns null (None segment selected back)

### LINKED ISSUE #282: Crash on setting target segment back to "None"
Accidentally discovered a bug (leading to a crash) in the EBP module:

When you set a target segment and then try to set it back to "None", Slicer crashes with this message:

_"GetSegmentCenter: Segment not found:"_

Image

Has anyone seen this before?

ferdymercury:
@LiBu981 does this PR fix the crash you found?

Closes https://github.com/SlicerRt/SlicerRT/issues/282

fyi @cpinter

### DIFF
--- a/Beams/MRML/vtkMRMLRTPlanNode.cxx
+++ b/Beams/MRML/vtkMRMLRTPlanNode.cxx
@@ -886,6 +886,11 @@ bool vtkMRMLRTPlanNode::ComputeTargetVolumeCenter(double center[3])
   }
 
   double* centerRas = segmentationNode->GetSegmentCenterRAS(this->TargetSegmentID);
+  if (!centerRas)
+  {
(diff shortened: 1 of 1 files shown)


LiBu981:
Yes, there is no crash anymore! Only some warnings:

{93692D05-84A5-4AD9-9881-14462AF8C75C}

cpinter:
What was the issue? I see that `this->TargetSegmentID` being null is handled. It was not null but empty string for example?

I wanted to fix this but bumped into the build error and then the startup crash :) (see "the PR")

ferdymercury:
> It was not null but empty string for example?

Was maybe "None"?

image

cpinter:
OK. I was asking because I was thinking we could catch this sooner. Anyway I think we can merge it because the important thing is to fix the crash. Thank you!

PR #286, opened 2025-11-10
FIX: Ensure creation of beam parameter tab in beam module by calling defineBeamParameters after engine registration

LiBu981:


ExternalBeamPlanning/Widgets/Python/DoseEngines/AbstractScriptedDoseEngine.py
```diff
@@ -44,3 +44,4 @@ def register(self):
     handler = qSlicerExternalBeamPlanningModuleWidgetsPythonQt.qSlicerDoseEnginePluginHandler()
     engineHandlerSingleton = handler.instance()
     engineHandlerSingleton.registerDoseEngine(self.scriptedEngine)
+    self.defineBeamParameters()
```
cpinter: Hm, you added a call, but I don't see the old one removed. Do you know who called this function before? Thanks!

LiBu981: I think from here:

```
//----------------------------------------------------------------------------
void qSlicerAbstractDoseEngine::registerBeamParametersTabWidget(qMRMLBeamParametersTabWidget* tabWidget)
{
  qSlicerAbstractDoseEngine::m_BeamParametersTabWidgets.insert(tabWidget);

  // Make sure beam parameters specified by engines are defined in the tab widget
  qSlicerDoseEnginePluginHandler::DoseEngineListType engines =
    qSlicerDoseEnginePluginHandler::instance()->registeredDoseEngines();
  for (qSlicerDoseEnginePluginHandler::DoseEngineListType::iterator engineIt = engines.begin(); engineIt != engines.end(); ++engineIt)
  {
    (*engineIt)->defineBeamParameters();
  }
}
```

called in

```
//-----------------------------------------------------------------------------
void qSlicerExternalBeamPlanningModuleWidget::setup()
{
   ...

  // Register the Beams module's beam parameters tab widget to the dose engines (do it here because of how the dependencies are set)
  qSlicerAbstractDoseEngine::registerBeamParametersTabWidget(qSlicerAbstractDoseEngine::beamParametersTabWidgetFromBeamsModule());

...
}
```

and _setup()_ is called before _enter()_, so the engines are not registered yet.

cpinter: Thanks! Well a proper solution would be to fix it in a way that there is no duplication. I think it would be a better solution to move this call from `setup` to `enter`, and remove the extra Python call.

LiBu981: Okay, I moved the registering of the beam parameters into `enter()` after registering the dose engines and removed the python call.

cpinter: Great, thanks! Please confirm that it works for you and we can merge. Thank you very much!

cpinter:
I ended up testing this and cleaning up the commits, so it's integrated in https://github.com/SlicerRt/SlicerRT/commit/e0b6873def9c3748b47cdb69b90e0c61ba2dbeec

Please note that
* There is one commit (in a PR it's always best to amend the one commit you have)
* The commit prefix is according to [Slicer style](https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages)

Issue #284, opened 2025-11-04
Crash when setting up scripted dose engines in External Beam Planning


cpinter:
The crash happens when we want to register the dose engines (which now happens on entering the module instead of startup):

https://github.com/SlicerRt/SlicerRT/blob/849f5e760de917424a4bb70943c8a5e1e08cccae/ExternalBeamPlanning/qSlicerExternalBeamPlanningModuleWidget.cxx#L217-L230

The crash itself happens here:

https://github.com/SlicerRt/SlicerRT/blob/849f5e760de917424a4bb70943c8a5e1e08cccae/ExternalBeamPlanning/Widgets/qSlicerScriptedDoseEngine.cxx#L137-L154

Basically what happens is that in the call at line 146 the `module` object becomes corrupted (parts of that memory seems to get overwritten), and then at line 150 the crash happens with access violation.

<img width="1198" height="366" alt="Image" src="https://github.com/user-attachments/assets/a8e2ab02-7d7e-4d7b-a7c3-27cc00c1ae85" />

The line where the `module` object gets corrupted is this:

https://github.com/Slicer/Slicer/blob/3c9e54198efe8aa116912e67a441b630ee836a71/Base/QTCore/qSlicerScriptedUtils.cxx#L52

To me it seems probable that the same PythonQt update that [broke the build](27e28dfa76f67a08dfcd6977f5fe0097ee32744a) is responsible for this crash. maybe the correct way of doing this has changed.

cpinter:
Some updates:

- [ ] The crash happens even if we copy-paste the snippet in the python console (not going through PythonQt)
- [ ] The Mock engine produces the crash too, not only the pyRadPlan engine
- [ ] There is no crash in Slicer 5.8.1, so some change after that must be related

More ideas (by @lassoan ):
- [ ] It would be worth checking who is the owner of the engine object. There is only `slicer.modules.doseEngineNames`, but it is not ownership (the created dose engine could be deleted unless saved in some global python variable). Maybe `register()` saves it, but need to check. For example we could try if we save the instantiated engine object in a variable `slicer.modules.doseEngines` for example, if it helps

lassoan:
🎉

PR #288, opened 2025-11-11
BUG: DoseEngines and PlanOptimizers now import correctly

xskere:
`ExternalBeamPlanning/Widgets/qSlicerScriptedDoseEngine.cxx` `ExternalBeamPlanning/Widgets/qSlicerScriptedPlanOptimizer.cxx`

Both of these files had a problem in their `setPythonSource` method. The methods attempted to import the modules without using the full module name. For example, when importing the `MockDoseEngine`, it was being imported as: `MockDoseEngine`. However, the correct way would be to import: `DoseEngines.MockDoseEngine`.

Same thing happened with the plan optimizers, but instead of adding `DoseEngines.` to the module import, it should be `PlanOptimizers.`

Re #284

### DIFF
--- a/ExternalBeamPlanning/Widgets/qSlicerScriptedDoseEngine.cxx
+++ b/ExternalBeamPlanning/Widgets/qSlicerScriptedDoseEngine.cxx
@@ -130,8 +130,11 @@ bool qSlicerScriptedDoseEngine::setPythonSource(const QString newPythonSource)
   PyObject * main_module = PyImport_AddModule("__main__");
   PyObject * global_dict = PyModule_GetDict(main_module);
 
+  // Prefix module name with "DoseEngines." for correct Python module import, without it the import would not work
+  QByteArray fullModuleName = "DoseEngines." + moduleName.toUtf8();
+
   // Get a reference (or create if needed) the <moduleName> python module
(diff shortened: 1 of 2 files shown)


ExternalBeamPlanning/Widgets/qSlicerScriptedDoseEngine.cxx
```diff
@@ -130,8 +130,10 @@ bool qSlicerScriptedDoseEngine::setPythonSource(const QString newPythonSource)
   PyObject * main_module = PyImport_AddModule("__main__");
   PyObject * global_dict = PyModule_GetDict(main_module);
 
+  QByteArray fullModuleName = "DoseEngines." + moduleName.toUtf8();
```
cpinter, line 134: Since this line is a simple fix for a complicated error, I think it's worth adding a comment about it. In the other class as well. Thanks.

cpinter (changes requested):
Just one small request, thanks!

ExternalBeamPlanning/Widgets/qSlicerScriptedDoseEngine.cxx
xskere: Both comments added!

cpinter (approved):

Issue #291, opened 2025-12-04
Include the IPOPT library in the superbuild


cpinter:
Modify the CMakeList of SlicerRT so that it also builds and links the
"IPOPT" library (https://github.com/coin-or/Ipopt) via FetchContent or
so. (See also
https://stackoverflow.com/questions/72027758/install-ipopt-locally-into-c-codebase-with-cmake-externalproject).

cpinter:
From @xskere:

There are several issues when dealing with adding Ipopt to SlicerRT.
 
The main repository for Ipopt does not include a way to build it with CMake. It uses a very long custom .sh file or uses Coinbrew (a custom program to allow easier installation), which is basically the same.
 
In the [stack overflow question](https://stackoverflow.com/questions/72027758/install-ipopt-locally-into-c-codebase-with-cmake-externalproject) sent with the task they seemed to have the same problem and no definitive answer was given.
 
After realizing that there was a need to allow CMake building for Ipopt I checked to see if there was already work being done for this, and found this repository:  [https://github.com/rjodon/coinor-ipopt-with-cmake](https://github.com/rjodon/coinor-ipopt-with-cmake). It is quite old, but as I have looked around this is the only example of Ipopt being built with CMake.
 
Apart from this, Ipopt uses a lot of Third Party software, and needs at least one of the following solver for systems of linear equations:
 
- MUMPS (MUltifrontal Massively Parallel sparse direct Solver) to build this a Fortran compiler is needed.
- MA27, MA57, HSL_MA77, HSL_MA86, or HSL_MA97 from the [Harwell Subroutines Library](http://hsl.rl.ac.uk/) (HSL).
- [Parallel Sparse Direct Linear Solver](https://panua.ch/pardiso/)  (Pardiso).
- [Sparse Parallel Robust Algorithms Library](https://github.com/ralna/spral)  (SPRAL).
- Watson Sparse Matrix Package.
 
Other software required for Ipopt to work, are the following:
 
- BLAS (Basic Linear Algebra Subroutines)
- LAPACK (Linear Algebra PACKage).

@ferdymercury Please let us know what you think should be next with this, or if you'd like to meet and discuss

ferdymercury:
I think on OSX, IPOPT can be installed using Homebrew. For Windows there are binary releases, but again Windows can be skipped.

For Linux, for the Third Party MUMPS, is using gfortran ok?.

HSL would be also beneficial:
[...]

PR #295, opened 2025-12-05
ENH: Ipopt added to superbuild

xskere:
- Added support for Ipopt library
- Added support for MUMPS linear solver dependency
- Added support for HSL linear solver dependency: 
    Requires manual download due to licensing (free for academic use) 
    Download from: https://licences.stfc.ac.uk/product/coin-hsl 
    Extract to 'coinhsl' directory in HSL source tree HSL provides MA27, MA57, MA77, MA86, MA97 solvers

Re #291

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,7 @@ set(EXTENSION_SCREENSHOTURLS "https://www.slicer.org/slicerWiki/images/8/87/Slic
 set(EXTENSION_STATUS "Beta")
 set(EXTENSION_DEPENDS "NA") # Specified as a space separated list or 'NA' if any
 set(EXTENSION_BUILD_SUBDIRECTORY inner-build)
+set(EXTENSION_BUILDS_IPOPT OFF)
 
 set(SUPERBUILD_TOPLEVEL_PROJECT inner)
 
@@ -52,6 +53,12 @@ endif()
 # Extension dependencies
 find_package(Slicer COMPONENTS ConfigurePrerequisites REQUIRED)
 project(SlicerRT)
+
+# Enable Fortran for MUMPS/LAPACK support in external projects (if optimization is enabled)
+if(EXTENSION_BUILDS_IPOPT AND UNIX)
+  enable_language(Fortran)
+endif()
+
 find_package(Slicer REQUIRED)
 include(${Slicer_USE_FILE})
 mark_as_superbuild(Slicer_DIR)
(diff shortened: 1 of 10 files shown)


ferdymercury:
Thanks a lot.

One question to @xskere: what is the main reason why Ipopt is not easily compatible with a CMake build on Windows? Is it because of the absence of a compatible Fortran compiler? Or alternative grabbing the [Windows binary release](https://github.com/coin-or/Ipopt/releases) directly might lead to incompatibilities?

Current solution in this PR is perfectly fine to me since I do not use Windows, but @lassoan insisted in the associated issue that we should try to make it compatible.

What I wanted to know more is whether "it's impossible that will ever be compatible because XYZ" or rather "setting up the CMake build compatible for Windows is possible but will take WWW hours, so skipping for the moment".

Another point @lassoan mentioned is the potential addition of Pardiso MKL (and potentially Lapack/Blas through it) by copy-pasting the relevant lines from https://github.com/febiosoftware/FEBio/blob/develop/CMakeLists.txt#L334 Maybe this can be moved a future issue/improvement if it's hard to add at the current stage.

xskere:
@ferdymercury 

Sorry about the late reply!

After taking another deep look through this, these are the main problems I have found:

CMake build on Windows is rather difficult because in the first place Ipopt does not support this.  

Even after installing different Fortran compilers such as ifx, that comes with Intel's OneApi (also adding MKL, Lapack/Blas), and gfortran (MSYS2) and using the provided branches that made Ipopt use CMake the builds still fail.

I do not think it's an impossible endeavor, but I do believe it will take a lot of time to fix this problem.

ferdymercury:
Thanks for the feedback.

One follow-up question:

For Windows, wouldn't it be possible to download the DLL files from https://github.com/coin-or/Ipopt/releases/download/releases%2F3.14.19/Ipopt-3.14.19-win64-msvs2022-md.zip and then just use the API (headers)?

xskere:
> For Windows, wouldn't it be possible to download the DLL files from https://github.com/coin-or/Ipopt/releases/download/releases%2F3.14.19/Ipopt-3.14.19-win64-msvs2022-md.zip and then just use the API (headers)?

Yes! I think it would be possible as a workaround, I will begin work on this now and update this PR so that windows builds with this instead of not building at all

lassoan:
In general, we cannot let a library to dictate exact compiler version, compilation options, etc. because to build an application we need to integrate hundreds of libraries. We build Slicer with v143 toolset and compatibility within v143 versions have been fairly good, so there may be a chance that these ipopt binaries would work, but it would be quite fragile.

Would it be possible to use the ipopt executable instead of linking DLLs?

ferdymercury:
> Would it be possible to use the ipopt executable

Not really, because the problems solved by IPopt in inverse radiotherapy treatment planning have a size of 'GB' and there is no way that I know that we can efficiently forward the information on the C++ side living in SlicerRT scene to the executable.

xskere:
I have added the Ipopt test and it successfully runs on Linux (I still need to test for Windows as it's done with the binaries and everything is different)

image


I commented out everything related to HSL because I do not have a license, but everything should work if it is set up with a license.

xskere:
I have also tested for Windows and it works with MUMPS as well!

ferdymercury:
awesome! LGTM to merge!! thanks

PR #297, opened 2026-01-04
Moved documentation from the old wiki

lassoan:
https://www.slicer.org/w/index.php/Documentation/Nightly/Extensions/SlicerRT

The documentation was not improved, merely reorganized and reformatted. Images were retrieved from the wiki as much as it was possible (some images were corrupted and only the thumbnail was available).


cpinter:
This is excellent, @lassoan ! Thanks a lot for working on this, a much needed improvement!

Issue #298, opened 2026-01-15
Create IPOPT optimizer class


xskere:
Make use of the IPOPT-C++ API by porting the example of https://github.com/e0404/pyRadPlan/blob/main/pyRadPlan/optimization/solvers/_ipopt.py or https://github.com/e0404/matRad/blob/master/matRad/optimization/optimizer/matRad_OptimizerIPOPT.m to a native C++ class within SlicerRT using directly the IPOPT C++ API.

cpinter:
@ferdymercury would like to include IPOPT in the packages distributed in the extensions manager.
@xskere checked the package sizes, and this is the result:
With IPOPT: 62.2M
Without IPOPT: 22.3M

The difference is basically three-fold.

@ferdymercury Do you definitely need to have this in Extensions Manager? I don't know how many users need this library, but I'd argue it's probably quite low. Tripling the package size to include this seems exaggarated to me.
@lassoan What do you think?

ferdymercury:
Thanks!

Can this be an opt-in like pyradplan ? Meaning, when you go to external Beams, it asks you if you want to install pyradplan. Could it ask the same for IPOPT ? Meaning it could be a separate extension that could be added if prompted?

cpinter:
I think it's a good idea to separate it into its extension and allow the user to opt in!

ferdymercury:
> I think it's a good idea to separate it into its extension and allow the user to opt in!

Sounds good !

PR #299, opened 2026-01-21
ENH: Added qSlicerIpoptOptimizer and test

xskere:
A new class called qSlicerIpoptOptimizer was added to the ExternalBeamPlanning Widgets directory.

This class functions as an interface from which to use IPOPT. Also, a new test was added to the ExternalBeamPlanning module to verify that the class does indeed work, which it did:

```
...
44: ✓ All tests passed successfully!
1/1 Test #44: qSlicerIpoptOptimizerBasicTest ...   Passed    1.62 sec

The following tests passed:
	qSlicerIpoptOptimizerBasicTest

100% tests passed, 0 tests failed out of 1

Label Time Summary:
ExternalBeamPlanning    =   1.62 sec*proc (1 test)
IPOPT                   =   1.62 sec*proc (1 test)
Optimization            =   1.62 sec*proc (1 test)

Total Test time (real) =   1.62 sec
```

This class was created after: https://github.com/e0404/pyRadPlan/blob/main/pyRadPlan/optimization/solvers/_ipopt.py
However, some modifications were made to make it work in C++, while maintaining the functionality of the class mentioned above.

Re #298

### DIFF
--- a/ExternalBeamPlanning/Testing/Cxx/qSlicerIpoptOptimizerTest.cxx
+++ b/ExternalBeamPlanning/Testing/Cxx/qSlicerIpoptOptimizerTest.cxx
@@ -1,0 +1,116 @@
+#include "../../Widgets/qSlicerIpoptOptimizer.h"
+
+#include <QCoreApplication>
+#include <QDebug>
+#include <iostream>
+#include <cmath>
+
+int main(int argc, char* argv[])
+{
+    QCoreApplication app(argc, argv);
+
+    std::cout << "Testing qSlicerIpoptOptimizer..." << std::endl;
+
+    // Create optimizer
+    qSlicerIpoptOptimizer optimizer;
+
+    // Test 1: Simple quadratic function - minimize (x-2)^2 + (y-1)^2
+    std::cout << "\n=== Test 1: Simple Quadratic Optimization ===" << std::endl;
+
+    auto objective = [](const qSlicerIpoptOptimizer::Array& x) -> double {
+        return (x[0] - 2.0) * (x[0] - 2.0) + (x[1] - 1.0) * (x[1] - 1.0);
+    };
+
+    auto gradient = [](const qSlicerIpoptOptimizer::Array& x) -> qSlicerIpoptOptimizer::Array {
+        qSlicerIpoptOptimizer::Array grad(2);
+        grad[0] = 2.0 * (x[0] - 2.0);  // d/dx
+        grad[1] = 2.0 * (x[1] - 1.0);  // d/dy
+        return grad;
+    };
+
+    optimizer.setObjectiveFunction(objective);
+    optimizer.setGradientFunction(gradient);
+
+    // Set stricter tolerance for testing
+    optimizer.setAbsoluteObjectiveTolerance(1e-8);
+    optimizer.setMaxIterations(100);
+
+    // Starting point
+    qSlicerIpoptOptimizer::Array x0 = {0.0, 0.0};
+
+    // Solve
+    auto result = optimizer.solveProblem(x0);
+
+    // Check results
+    if (result.success) {
+        std::cout << "✓ Optimization succeeded!" << std::endl;
+        std::cout << "  Solution: [" << result.solution[0] << ", " << result.solution[1] << "]" << std::endl;
+        std::cout << "  Expected: [2.0, 1.0]" << std::endl;
+        std::cout << "  Final objective: " << result.final_objective_value << std::endl;
+        std::cout << "  Iterations: " << result.iteration_count << std::endl;
+
+        // Verify solution accuracy
+        double error_x = std::abs(result.solution[0] - 2.0);
+        double error_y = std::abs(result.solution[1] - 1.0);
+        double max_error = std::max(error_x, error_y);
+
+        if (max_error < 1e-6) {
+            std::cout << "✓ Solution accuracy test passed (error: " << max_error << ")" << std::endl;
+        } else {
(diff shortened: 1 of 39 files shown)


ferdymercury:
Thanks a lot!
Would it be doable to also implement "constraints", not only "objectives" ?

As done in:
https://github.com/e0404/matRad/blob/4f277bb66060343e06424e2ce5eb202c50534a9d/matRad/optimization/optimizer/matRad_OptimizerIPOPT.m#L136
and
https://github.com/e0404/matRad/blob/4f277bb66060343e06424e2ce5eb202c50534a9d/matRad/optimization/optimizer/matRad_OptimizerIPOPT.m#L141

I guess it will need an additional table above "Objectives" in the Slicer GUI called "Constraints".
It can be a different PR though if you prefer.

xskere:
IPOPT now has been integrated as an optimizer for External Beam Planning with both objectives and constraints.

The objectives and constraints have been added following the matRad implementation ([objectives](https://github.com/e0404/matRad/tree/4f277bb66060343e06424e2ce5eb202c50534a9d/matRad/optimization/%2BDoseObjectives), [constraints](https://github.com/e0404/matRad/tree/4f277bb66060343e06424e2ce5eb202c50534a9d/matRad/optimization/%2BDoseConstraints)).

To show both objectives and constraints on the table without mixing them completely, a checkbox has been added to the objectives table, which if checked means that instead of an objective, it's a constraint (this column is only visible if the optimizer selected has constraints implemented)

<div align='center'>
image
</div>

ferdymercury:
Looks awesome, thanks a lot.

What happens if I choose eg objective SquaredOverdosing, and then click on "Constraint" checkbox. Does it automatically unselect the objective dropdown menu? Since that objective does not exist as constraint function.

What if I select MaxDVH as objective, and then click on Constraint checkbox. Does it map it to MinMaxDVH constraint with min=0 (or -inf) and max= the old-max I had set up for objective?

ferdymercury:
Minor note wrt alphabetical ordering: It might be confusing if maxDose appears before minDose in Parameters

xskere:
> What if I select MaxDVH as objective, and then click on Constraint checkbox. Does it map it to MinMaxDVH constraint with min=0 (or -inf) and max= the old-max I had set up for objective?

I think this can be done (and I will add so if you would prefer it to be like that), but as of now what it does it's simply reset to the objective/constraint as if it were added anew. It does not change it dynamically between them.

ExternalBeamPlanning/Widgets/qSlicerIpoptOptimizer.h
```diff
@@ -0,0 +1,159 @@
...
+public:
+  using Array = std::vector<double>;
+  using ObjectiveFunction = std::function<double(const Array&)>;
+  using GradientFunction = std::function<Array(const Array&)>;
+
+  struct Options {
+    int print_level = 5;
+    QString print_user_options = "no";
+    QString print_options_documentation = "no";
+    double tol = 1e-10;
```
ferdymercury, line 71: It would be beneficial if one could set these options from the GUI, from an Advanced Settings foldable menu

xskere: Yeah, I planned to but didn't know where or how, I added them as a foldable menu, these 'IPOPT Advanced Settings' show only if IPOPT is selected as the optimizer.

ExternalBeamPlanning/Testing/Cxx/qSlicerIpoptOptimizerTest.cxx
```diff
@@ -0,0 +1,116 @@
...
+            std::cout << "✓ Rosenbrock accuracy test passed (error: " << max_error << ")" << std::endl;
+        } else {
+            std::cout << "✗ Rosenbrock accuracy test failed (error: " << max_error << ")" << std::endl;
+            return 1;
+        }
+    } else {
+        std::cout << "✗ Rosenbrock optimization failed with status: " << result_rb.status << std::endl;
+        return 1;
+    }
+
```
ferdymercury, line 114:
```suggestion
    // For a more advanced interactive test with proton patients, please check out https://github.com/SebastiaanBreedveld/TROTS/tree/master/SlicerRT_import
```

It would be helpful to have the python script that imports the Dij matrices into SlicerRT logged somewhere, either here or by submitting a PR to https://github.com/SebastiaanBreedveld/TROTS

That way, even if it's not a real ctest, anyone in the future could cross-check easily via the script and then the GUI, that things are still working reasonably well after this or that update.

(Not the MATLAB part, just the part that loads the Dij and runs the optimizer, which I believe was done via the the Python-SlicerRT scripting interface?)

xskere: Added! I also submitted the PR to https://github.com/SebastiaanBreedveld/TROTS to add the comparison script!

xskere: If no matlab weights are present, it does not take that into consideration, but if present it adds them to the comparison

ferdymercury: Awesome, thanks!

ferdymercury (approved):
Thanks!!

cpinter:
@ferdymercury I can merge this if you agree. Thanks!

ferdymercury:
Great, thanks!

PR #303, opened 2026-02-13
BUG: Replace triangulation-based contour-to-surface with rasterization approach

melandur:
## Summary

Replace the dynamic programming triangulation in `vtkPlanarContourToClosedSurfaceConversionRule::Convert()` with a rasterization-based algorithm that produces artifact-free closed surfaces from DICOM RT Structure Set (RTSS) planar contours.

### Problem

The previous triangulation approach (`TriangulateBetweenContours` with `Branch`/`EndCapping`) suffered from:
- **Overshooting artifacts** where surfaces extended beyond actual contour boundaries when contour shapes changed between slices
- **Vertical sheet artifacts** from incorrect point correspondence in the dynamic programming table
- **Blocky/jagged 2D cross-sections** because the 2D slice views intersect the broken 3D surface

### Solution

The new algorithm uses a rasterization + marching cubes approach:

1. **Group contours by Z-plane** with tolerance-based co-planar detection
2. **Rasterize each contour group** into a 2D binary slice using `vtkPolyDataToImageStencil` (even-odd fill correctly handles nested contours / holes)
3. **Build an isotropic float volume** with linear interpolation between adjacent contour slices (shape-based interpolation), creating smooth transitions between differently-shaped contours
4. **Extract the closed surface** using `vtkMarchingCubes` at the 0.5 isovalue
5. **Smooth and clean** with `vtkWindowedSincPolyDataFilter` and `vtkCleanPolyData`

This approach is the standard method used in radiation therapy software for contour-to-surface conversion and eliminates the class of artifacts inherent to direct inter-slice triangulation.

### Additional changes

Defensive improvements to retained helper functions:
- Default slice thickness changed from 0.0 to 2.0 mm (prevents degenerate geometry for single-slice structures)
- Minimum fallback in `GetSpacingBetweenLines` when thickness parameter is zero
- Guard in `EndCapping` against near-zero spacing
- Fix `GetNumberOfLinesOnPlane` co-planar threshold minimum
- Fix copy-paste bug in `TriangulateBetweenContours` (`numberOfPointsInLine2` vs `numberOfPointsInLine1`)

## Test plan

- [x] Load DICOM RTSS file with multi-slice structures - verify smooth 3D surfaces without overshooting
- [x] Verify 2D axial slice view shows proper filled contours
- [x] Verify structures with nested contours (holes) render correctly
- [x] Verify single-slice structures do not produce degenerate geometry

### DIFF
--- a/DicomRtImportExport/ConversionRules/vtkPlanarContourToClosedSurfaceConversionRule.cxx
+++ b/DicomRtImportExport/ConversionRules/vtkPlanarContourToClosedSurfaceConversionRule.cxx
@@ -28,6 +28,7 @@
 #include <vtkImageDilateErode3D.h>
 #include <vtkImageStencil.h>
 #include <vtkLine.h>
+#include <vtkMarchingCubes.h>
 #include <vtkMarchingSquares.h>
 #include <vtkPlane.h>
 #include <vtkPolyDataToImageStencil.h>
@@ -39,12 +40,15 @@
 #include <vtkTransformPolyDataFilter.h>
 #include <vtkUnstructuredGrid.h>
 #include <vtkVersion.h> // For VTK_VERSION_*
+#include <vtkWindowedSincPolyDataFilter.h>
 
 // vtkAddon includes
 #include <vtkAddonMathUtilities.h>
 
 // STD includes
 #include <algorithm>
+#include <cmath>
+#include <cstring>
 
 // SegmentationCore includes
 #if Slicer_VERSION_MAJOR >= 5 || (Slicer_VERSION_MAJOR >= 4 && Slicer_VERSION_MINOR >= 11)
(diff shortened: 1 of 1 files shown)


lassoan:
Conversion from contour directly to closed surface has several advantages over simple rasterization (most importantly, it is a lossless operation, while rasterization is not; it also naturally handles contours with varying distance; its memory usage and comptutation time is near-zero). Overshooting is needed (flat top is not real). I'm not sure what you mean by vertical sheet artifacts and blocky cross-sections, of course bugs are always possible, but it may be just rendering artifact (closed surface rendering in slice views has a known bug).

That said, rasterization is a valid approach, too, and if you have use cases where it is preferable then you are welcome to contribute it. You don't need to modify the existing `vtkPlanarContourToClosedSurfaceByRasterizationRule` rule, but you can simply add a new `vtkPlanarContourToClosedSurfaceByRasterizationConversionRule`. The converter will chose the rule that has the lower cost by default, so I would recommend to keep the cost higher than `vtkPlanarContourToClosedSurfaceByRasterizationRule` to not change the default conversion path for now.

Issue #306, opened 2026-03-03
Unused beam geometry and MLC parameters remain editable for inverse planning


LiBu981:
**Current behavior:**
When a plan is set to "inverse", some beam parameters should be made unavailable in the UI.
For example for the _pyRadPlan_ dose engine only the gantry and couch angle are currently used, making it confusing for the user that other parameters remain editable.

**Expected behavior:**
These unused parameters should be disabled on choosing inverse planning (or should this be adjustable for different dose engines?).

PR #307, opened 2026-03-03
ENH: Disable unused beam geometry parameters and MLC tab for inverse planning

LiBu981:
### Problem:
Currently, when a plan is set to use inverse planning, unused beam geometry parameters and MLC parameters remain editable in the Beam Module. This can lead to user confusion, as these values are not currently used by inverse planning dose engines (e.g. _pyRadPlan_ dose engine).

### Solution:

- Added an `InverseFlag `to `vtkMRMLRTPlanNode` to track whether inverse planning is active.
- Updated `qMRMLBeamParametersTabWidget::updateWidgetFromMRML() ` to enable/disable beam geometry and MLC widgets based on the current plan’s inverse flag.
- Ensured that widgets are always explicitly enabled/disabled depending on the inverse flag, preventing stale UI state when switching plans.

### Behavior Changes:

- Beam geometry and MLC widgets are now grayed out when using inverse plans.
- Switching between inverse and non-inverse plans correctly restores the enabled state.

### Remaining Questions:
- Currently explicitly enabling/disabling SAD, jaws, collimator angle and MLC tab. Are there other parameters to include here?
- Should the available parameters be dependent on the selected dose engine?

### See issue:
[issue 306](https://github.com/SlicerRt/SlicerRT/issues/306)

### DIFF
--- a/Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
+++ b/Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
@@ -404,22 +404,13 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
 
   // For inverse plans
   vtkMRMLRTPlanNode* planNode = d->BeamNode->GetParentPlanNode();
-  if (planNode->GetInverseFlag())
-  {
-    d->doubleSpinBox_SAD->setEnabled(false);
-    d->RangeWidget_XJawsPosition->setEnabled(false);
-    d->RangeWidget_YJawsPosition->setEnabled(false);
-    d->SliderWidget_CollimatorAngle->setEnabled(false);
-    d->tabMultiLeafCollimator->setEnabled(false);
-  }
-  else
-  {
-    d->doubleSpinBox_SAD->setEnabled(true);
-    d->RangeWidget_XJawsPosition->setEnabled(true);
-    d->RangeWidget_YJawsPosition->setEnabled(true);
-    d->SliderWidget_CollimatorAngle->setEnabled(true);
-    d->tabMultiLeafCollimator->setEnabled(true);
-  }
+  bool inversePlan = planNode->GetInverseFlag();
+
+  d->doubleSpinBox_SAD->setEnabled(!inversePlan);
+  d->RangeWidget_XJawsPosition->setEnabled(!inversePlan);
+  d->RangeWidget_YJawsPosition->setEnabled(!inversePlan);
+  d->SliderWidget_CollimatorAngle->setEnabled(!inversePlan);
+  d->tabMultiLeafCollimator->setEnabled(!inversePlan);
 }
 
 //-----------------------------------------------------------------------------
--- a/Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
+++ b/Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
@@ -401,6 +401,25 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
       d->CollapsibleButton_RangeShifterParameters->setEnabled(false);
     }
   }
+
+  // For inverse plans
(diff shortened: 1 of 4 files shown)


Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
```diff
@@ -401,6 +401,25 @@ void qMRMLBeamParametersTabWidget::updateWidgetFromMRML()
...
+    d->tabMultiLeafCollimator->setEnabled(false);
+  }
+  else
+  {
+    d->doubleSpinBox_SAD->setEnabled(true);
+    d->RangeWidget_XJawsPosition->setEnabled(true);
+    d->RangeWidget_YJawsPosition->setEnabled(true);
+    d->SliderWidget_CollimatorAngle->setEnabled(true);
+    d->tabMultiLeafCollimator->setEnabled(true);
+  }
```
cpinter:
```suggestion
  vtkMRMLRTPlanNode* planNode = d->BeamNode->GetParentPlanNode();
  bool inversePlan = planNode->GetInverseFlag();
  
  d->doubleSpinBox_SAD->setEnabled(!inversePlan);
  d->RangeWidget_XJawsPosition->setEnabled(!inversePlan);
  d->RangeWidget_YJawsPosition->setEnabled(!inversePlan);
  d->SliderWidget_CollimatorAngle->setEnabled(!inversePlan);
  d->tabMultiLeafCollimator->setEnabled(!inversePlan);
```

ExternalBeamPlanning/qSlicerExternalBeamPlanningModuleWidget.cxx
```diff
@@ -1000,24 +1003,31 @@ void qSlicerExternalBeamPlanningModuleWidget::inversePlanningCheckboxStateChange
...
-  // Update Optimization engines
-  this->updatePlanOptimizers();
+  vtkMRMLRTPlanNode* planNode = vtkMRMLRTPlanNode::SafeDownCast(d->MRMLNodeComboBox_RtPlan->currentNode());
+  if (!planNode)
+  {
+    qCritical() << Q_FUNC_INFO << ": Invalid RT plan node";
+    return;
+  }
 
   if (d->checkBox_InversePlanning->isChecked())
```
cpinter: I think this also could be simplified as the above suggestion

cpinter:
Looks good to me, thank you! Just two small suggestions to simplify code.

Beams/Widgets/qMRMLBeamParametersTabWidget.cxx
LiBu981: Makes sense. Thank you! (See [commit 8176a9c](https://github.com/SlicerRt/SlicerRT/pull/307/changes/8176a9cb85830f5064d3a2da3c6d85d4454d25a5))

ExternalBeamPlanning/qSlicerExternalBeamPlanningModuleWidget.cxx
LiBu981: See [commit 8176a9c](https://github.com/SlicerRt/SlicerRT/pull/307/changes/8176a9cb85830f5064d3a2da3c6d85d4454d25a5)

LiBu981:
I noticed that the beam module was not updating on changing the inverse plan flag state, so I connected it through `vtkMRMLRTPlanNode::InversePlanFlagChanged` (see [commit 2b3c004](https://github.com/SlicerRt/SlicerRT/pull/307/changes/2b3c0045cf022ffd381f931c9ef64fa48250be22)). Is this correct?

cpinter:
I think it is. It might be an overkill defining a new event for that, but I think it's not worth the trouble investigating so I think we can keep it. Thank you!

cpinter:
Thank you!

I haven't tested this thoroughly as it seemed a relatively simple fix and all this inverse planning is kind of work in progress. So as the code looked good I decided to integrate it. If we find more issues we'll address them separately

PR #314, opened 2026-04-10
Iostream missing include fixes

sjh26:
- Adds iostream header where used
- Adds `std::` prefix to iostream function

See: https://discourse.vtk.org/t/important-update-standard-iostream-availability-from-vtk/16171


MichaelColonel:
I've tested this PR in Linux, and it compiles successfully after newly VTK 9.6 upgrade.

lassoan (approved):
The changes look good to me, thank you!

PR #315, opened 2026-04-21
ENH: Auto-create body segment and convert segmentation representation in pyRadPlan dose/optimization engine

LiBu981:
This PR adds two methods for the pyRadPlan dose and optimizer engines:

1. **Automatic body segment creation**: Without selected body segment, the user is prompted to create one from the current structure set. The new "BODY" segment is added directly as a binary labelmap to the segmentation.
(A body segment is required for pyRadPlan's optimizer for proper dose distribution, as dose is only placed in segments on which objectives were defined.)
2.  **Automatic segmentation representation conversion**: If the segmentation is not already in binary labelmap representation, it is automatically converted.
(Required for loading segments into python arrays via  `slicer.util.arrayFromSegmentBinaryLabelmap`.)
 (Adresses issue #316)

_Additional changes:_
- Per-beam dose  in pyRadPlanOptimizer is now loaded directly from pyRadPlan's optimized `result`
- Per-beam dose volume nodes are now directly converted to dose volumes
- Variable naming in pyRadPlan Python scripts corrected to camelCase

### DIFF
--- a/ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
+++ b/ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
@@ -16,6 +16,7 @@ class pyRadPlanEngine(AbstractScriptedDoseEngine):
     scriptedEngine.name = 'pyRadPlan'
     scriptedEngine.isInverse = True #pyRadPlan has Inverse planning capabilities, i.e., it can compute a dose influence matrix
     scriptedEngine.canDoIonPlan = True
+    scriptedEngine.supportsBodySegment = True
     AbstractScriptedDoseEngine.__init__(self, scriptedEngine)
 
   #------------------------------------------------------------------------------
@@ -35,15 +36,15 @@ class pyRadPlanEngine(AbstractScriptedDoseEngine):
   #------------------------------------------------------------------------------
   def updateBeamParametersForIonPlan(self, isIonPlanActive):
     if isIonPlanActive:
-      available_radiation_modes = ["protons", "carbons"]
-      parameter_label = "radiation mode (ion)"
+      availableRadiationModes = ["protons", "carbons"]
+      parameterLabel = "radiation mode (ion)"
     else:
-      available_radiation_modes = ["photons", "protons", "carbons"]
-      parameter_label = "radiation mode"
+      availableRadiationModes = ["photons", "protons", "carbons"]
+      parameterLabel = "radiation mode"
     
     self.scriptedEngine.updateBeamParameterComboBox(
-    "pyRadPlan parameters", "radiationMode", parameter_label,
-    "comment", available_radiation_modes, 0)
+    "pyRadPlan parameters", "radiationMode", parameterLabel,
+    "comment", availableRadiationModes, 0)
 
   #------------------------------------------------------------------------------
   def calculateDoseUsingEngine(self, beamNode, resultDoseVolumeNode):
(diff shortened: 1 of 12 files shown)


ExternalBeamPlanning/Widgets/Python/PyRadPlanUtils.py
```diff
@@ -17,8 +19,8 @@ def prepareCt(beamNode):
   return create_ct(cube_hu=image)
 
 #------------------------------------------------------------------------------
-def prepareCst(beamNode, ct):
-  from pyRadPlan.cst import StructureSet, create_voi
+def prepareCst(beamNode, ct, need_body=False):
```
cpinter: I'd prefer using only one style in one project. Please rename the new variables and methods in style of `need_body` to the style `needBody`. Thanks!

ExternalBeamPlanning/Widgets/Python/PyRadPlanUtils.py
```diff
@@ -27,20 +29,35 @@ def prepareCst(beamNode, ct):
   segmentationNode = planNode.GetSegmentationNode()
   segmentation = segmentationNode.GetSegmentation()
 
+  # TODO: should be handled with error messages in the UI instead of print statements in the console
+  if planNode.GetTargetSegmentID() == "":
+    print("Warning: No target segment selected in plan node. Please select a target segment and try again.")
```
cpinter: We have a better way of logging in Slicer. You already seem to import it. Please use `logging.warning` for these warnings (and error for errors, and info for general prints), everywhere. Unfortunately print only appears on the console, while the proper logging shows up in the log file as well, and it indicates severity (and can be filtered). Thanks!

ExternalBeamPlanning/Widgets/Python/PyRadPlanUtils.py
```diff
@@ -112,3 +129,92 @@ def preparePln(beamNode, ct, dose_grid_from_beamNode=False):
     pln["prop_opt"] = {"solver": "scipy"}
 
   return create_pln(pln)
+
+
+#------------------------------------------------------------------------------
+# HELPERS
+
+def getSegmentArrays(segmentationNode, referenceVolumeNode):
```
cpinter, line 135: Please add separators for all new functions, and add docstring explaining what they do.

cpinter (changes requested):
I added some minor comments.

Also I see that in this PR you have an older commit, and then the commit that seems to be the main one in the PR. Can you please clean this up?

Thanks a lot!

LiBu981:
@cpinter I squashed the commits and integrated your comments. I also extended the corrections for the logging to the pyRadPlan dose engine and optimizer scripts. Let me know if anything should be changed/improved :)

I also just realized that it might be a good idea to only enable the "Body volume" selection for engines which actually support using the body segment (so at the moment ony pyRadPlan). Otherwise it might be confusing for the user if it is selectable but not actually used in the engine's code.

image
image


Would you like me to include this?

cpinter:
> Would you like me to include this?

Yes please. Thank you :)

ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
```diff
@@ -7,6 +8,13 @@
 from DoseEngines import *
 from Python.PyRadPlanUtils import prepareCt, prepareCst, preparePln
 
+# Set up logging and print to python console in Slicer
```
cpinter: I don't understand this. Please just import logging and use it. See any Python file in Slicer.

ExternalBeamPlanning/Widgets/Python/PyRadPlanUtils.py
```diff
@@ -17,35 +26,47 @@ def prepareCt(beamNode):
...
-    segmentArray = slicer.util.arrayFromSegmentBinaryLabelmap(segmentationNode, id, referenceVolumeNode)
-    segmentArray = np.uint8(segmentArray)
-    segmentName = segmentation.GetSegment(id).GetName()
-    # Set VOI type as 'TARGET' if segment is selected as target, otherwise 'OAR'
-    voi_type = 'TARGET' if id==planNode.GetTargetSegmentID() else 'OAR'
+    segmentArray = segmentArrays[id]['array']
+    segmentName = segmentArrays[id]['name']
+    voi_type = 'OAR'
+    if id == planNode.GetBodySegmentID():
+      voi_type = 'EXTERNAL'
```
cpinter: Please stick to the camel case style (`voiType`)

ExternalBeamPlanning/Widgets/Python/DoseEngines/pyRadPlanEngine.py
LiBu981: This was for printing the log info also to the python console, but I removed it now.

LiBu981:
@cpinter I added the engine-specific en-/disabling of the body volume selection and changed, where possible, the naming to camel case.
As I needed to update thy pyRadPlan version anyway, I also updated how the optimized per-beam dose is loaded from the result in the the pyRadPlanOptimizer and added automatic conversion to dose volumes here.

Let me know if anything needs to be changed :)

cpinter:
Sorry for the delay; I don't see any potential issues, so merging it. I'll review the pyRadPlan files later, but it's your class anyway :)

Issue #313, opened 2026-04-10
Entering Beam module adds a new row of the selected dose engine parameters


xskere:
Each time the Beams module is entered, the specific parameters related to the selected dose engine are added again, even if they were added before:

<img width="514" height="467" alt="Image" src="https://github.com/user-attachments/assets/557c6eac-5fad-4fb4-b455-64305fe26962" />

This happens with every dose engine. If you enter the Beams module either directly or by clicking 'Edit' in a specific beam on External Beam Planning, another row of the parameters is added.

cpinter:
It is dangerous to create issues with no milestone. They get out of sight easily. Assign a milestone always please. If unsure, ask, if you cannot ask, decide (bugs normally in the current one, future plans up to decision)

LiBu981:
Does this solve the issue for you? #317

xskere:
> Does this solve the issue for you? [#317](https://github.com/SlicerRt/SlicerRT/pull/317)

Yes, it does, thank you!

cpinter:
Fix integrated as https://github.com/SlicerRt/SlicerRT/commit/1c5d2907b5585ed5309d0b0a826fca625c6540c4

PR #317, opened 2026-04-23
BUG: Fix duplicate dose-engine-specific beam parameters added on each EBP module entry

LiBu981:
This PR addresses Issue #313

`registerBeamParametersTabWidget()` is being called every time the External Beam Planning module is entered and re-adds the dose-engine-specific beam parameters to the tab widget, causing duplicate rows. Fixed by skipping registration if tab widget is already registered.


xskere:
Thank you again for this! I just tried it and it works completely fine now.

Issue #318, opened 2026-04-30
Multi-target support for RT plan


LiBu981:
Currently, the RT plan only supports a single target segment for dose calculation and isocenter placement. A multi-target feature was requested (for a student course) for treating multiple lesions in a single plan.

**Proposed approach:**
- Allow selecting multiple target segments in the EBP module's target segment selector widget (see below)
- Compute isocenter as the center of all selected target segments
- Dose engines have access to individual target segments (e.g needed to set all target segments as `TARGET` VOIs for the pyRadPlan dose engine)
- Merge selected segments into a combined labelmap for the planNode's `targetOrientedImageData `(e.g. needed for the Plastimach dose engine)

I have already implemented a working prototype of this feature and would be happy to share. Would this be a welcome addition?

<img width="500" height="300" alt="Image" src="https://github.com/user-attachments/assets/559afde5-dae8-4484-b9b4-66c14f4ec049" />

cpinter:
Please create a PR if there is no obstacle to do so, and we'll discuss there. In theory of course, it seems useful! Thank you!

LiBu981:
Finally got around to it. Just opened a PR :)

PR #319, opened 2026-05-28
ENH: Add multi-target segment support for RT plan isocenter and dose calculation

LiBu981:
Addresses issue #318.

_Open questions:_

- Saving multiple targets in one string (seperated by `|` ) acceptable?
(`planNode->SetTargetSegmentIDs(targetSegmentIDs.join("|").toUtf8().constData());`)

- How to handle isocenter from multiple targets?
Currently calculating mean of isocenters in `vtkMRMLRTPlanNode::ComputeTargetVolumeCenter` (`Beams/MRML/vtkMRMLRTPlanNode.cxx` lines 989-1035)

- How to remove shift between target segment & body segment selection widgets introduced by target multiSelection box? (see image)
image

### DIFF
--- a/Beams/MRML/vtkMRMLRTPlanNode.cxx
+++ b/Beams/MRML/vtkMRMLRTPlanNode.cxx
@@ -42,6 +42,10 @@
 #include <vtkSmartPointer.h>
 #include <vtkVariant.h>
 
+// Qt includes
+#include <QString>
+#include <QStringList>
+
 //------------------------------------------------------------------------------
 const char* vtkMRMLRTPlanNode::ISOCENTER_FIDUCIAL_NAME = "Isocenter";
 const int vtkMRMLRTPlanNode::ISOCENTER_FIDUCIAL_INDEX = 0;
@@ -66,7 +70,7 @@ vtkMRMLRTPlanNode::vtkMRMLRTPlanNode()
 //----------------------------------------------------------------------------
 vtkMRMLRTPlanNode::~vtkMRMLRTPlanNode()
 {
-  this->SetTargetSegmentID(nullptr);
+  this->SetTargetSegmentIDs(nullptr);
   this->SetBodySegmentID(nullptr);
   this->SetDoseEngineName(nullptr);
   this->SetPlanOptimizerName(nullptr);
@@ -80,7 +84,7 @@ void vtkMRMLRTPlanNode::WriteXML(ostream& of, int nIndent)
   // Write all MRML node attributes into output stream
   vtkMRMLWriteXMLBeginMacro(of);
   vtkMRMLWriteXMLIntMacro(NextBeamNumber, NextBeamNumber);
-  vtkMRMLWriteXMLStringMacro(TargetSegmentID, TargetSegmentID);
+  vtkMRMLWriteXMLStringMacro(TargetSegmentIDs, TargetSegmentIDs);
   vtkMRMLWriteXMLStringMacro(BodySegmentID, BodySegmentID);
   vtkMRMLWriteXMLStringMacro(DoseEngineName, DoseEngineName);
   vtkMRMLWriteXMLStringMacro(PlanOptimizerName, PlanOptimizerName);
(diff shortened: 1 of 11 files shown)


MichaelColonel:
> * How to remove shift between target segment & body segment selection widgets introduced by target multiSelection box? (see image)

I have changed a layout of these widgets with QGridLayout, so this shift must disappear.

That is the ExternalBeamPlanning/Resources/UI directory.
[UI.zip](https://github.com/user-attachments/files/28398141/UI.zip)

MichaelColonel:
> Saving multiple targets in one string (seperated by | ) acceptable?\n(planNode->SetTargetSegmentIDs(targetSegmentIDs.join(\"|\").toUtf8().constData());)

I think that a QStringList is more acceptable. Find and remove procedures are much easier as well. IMHO.

cpinter:
Thank you @LiBu981 !

I think manual serialization is not optimal, as @MichaelColonel says. You can use the MRML property for `std::vector<std::string>` (`vtkMRMLWriteXMLStdStringVectorMacro(xmlAttr, PropertyName, std::vector)` etc, see in `vtkMRMLNodePropertyMacros.h`).

I agree with using the mean of multiple isocenters (@ferdymercury what do you think?)

Remove the shift: see solution of @MichaelColonel , use if it you like it (I haven't checked). In general I think we can integrate this change even with this issue present. I can take a look at it afterwards from the main branch if needed.

ferdymercury:
> I agree with using the mean of multiple isocenters (@ferdymercury what do you think?)

Sounds good to me.

Can we still select different isocenters (or targets) for different beams, right?


A slightly related question: is there an easy button to get a 2D projection view where we see things from the beam eye view perspective, all targets projected? This way, if your scanning beam can cover an area of 30cm x 30cm and we can show that rectangle, we would easily see if the calculated mean isocenter of multiple targets and the associated beam-coverage-rectangle encloses all segmented targets or not.

cpinter:
There is a Beam's eye view button in the Room's eye view module. Not sure if it works without loading the machine. In any case the code is there so if we find a good place to put another button in EBP or Beams, we can do it easily.

LiBu981:
> I think manual serialization is not optimal, as @MichaelColonel says. You can use the MRML property for `std::vector<std::string>` (`vtkMRMLWriteXMLStdStringVectorMacro(xmlAttr, PropertyName, std::vector)` etc, see in `vtkMRMLNodePropertyMacros.h`).

@cpinter I changed the way the target segments are stored (see [commit a1b6a12](https://github.com/SlicerRt/SlicerRT/pull/319/changes/a1b6a127e355e4b0130a6ffecf17ba39bf35677d)). Let me know if there are any issues remaining with this.

> Remove the shift: see solution of @MichaelColonel , use if it you like it (I haven't checked). In general I think we can integrate this change even with this issue present. I can take a look at it afterwards from the main branch if needed.

Unfortunately, @MichaelColonel 's solution still didn't remove the issue for me, as the shift is still present when the multiselection box for the target segments is activated (when selecting a segmentation). After some more investigation, I found the real issue in Slicer's `qMRMLSegmentSelectorWidget.ui`. The `CheckableComboBox_Segment` (for multiselection) and the `comboBox_Segment` (without multiselection activated) are in different columns. I already tried setting the same column here, and it works perfectly!

image

LiBu981:
> There is a Beam's eye view button in the Room's eye view module. Not sure if it works without loading the machine. In any case the code is there so if we find a good place to put another button in EBP or Beams, we can do it easily.

I tried out including a BEV button in the EBP module's BeamTableView, by copying the `qSlicerRoomsEyeViewModuleWidget::onBeamsEyeViewButtonClicked()` and removing the call to set the machine parts opacity. @cpinter is this what you were thinking of? (See images below.) If yes, I can add it here or open a new PR.

Screenshot 2026-06-09 165624

Screenshot 2026-06-09 165654

Screenshot 2026-06-09 165707

cpinter:
Yes, perfect! Thank you @LiBu981 !

I think it would be great consolidating the BEV code to remove the duplication, but since @xskere has just been working on this, I think we should merge this first and let him simplify it afterwards.

I think you can add the BEV change here as a new commit. Thanks!

LiBu981:
The last commit integrates the BEV buttons. Let me know if anything needs to be changed here :)
I've also opened a PR in Slicer to fix the misalignment of the target and body volume dropdown boxes.

cpinter:
Thank you very much!

ferdymercury:
> The last commit integrates the BEV buttons. Let me know if anything needs to be changed here :)

Thanks a lot for this!!

A potential improvement would be that, when you click on BEV, the 3D camera switches to ortographic rather than perspective, since what you usually want to see is the projection of the beam extent into the contours to see if it fully covers the target.

PR #324, opened 2026-06-08
ENH: Restore machine parts opacity in Rooms Eye View when the viewpoint changes

xskere:
Opacity is lowered when entering the beam's eye viewpoint so the beam path can be seen through the machine. It was only restored again when the beam selection or the gantry angle changed.

Now it is also restored when the collimator or patient support angle changes, the other two angle-type controls.

It also restores the opacity as soon as the user manually moves the view away from the positioned viewpoint, by rotating, panning, or zooming, since the lowered opacity no longer makes sense once the beam's eye view is no longer being shown. This is detected through the signal the 3D view sends specifically for direct user interaction with the camera, which is not raised when the viewpoint is set up by the button itself, so positioning the view does not immediately undo its own opacity change.


cpinter:
Looks good, thank you!

PR #327, opened 2026-06-22
BUG: Fix camera and view detection for Beam's Eye View on External Beam Planning Module

xskere:
This fixes the same bug that happened on Room's Eye View that was fixed on commit https://github.com/SlicerRt/SlicerRT/commit/80487992f155544e5b0730d1519ecb174ddc9f01 (merged on https://github.com/SlicerRt/SlicerRT/pull/323).

When there were multiple views added by different extensions, the Beam's Eye View in the External Beam Planning module would pick up those views instead of main active, making it seem as if clicking the button did nothing.

Also: fixed a bug in which a macro being called with too many arguments caused a crash when building with GCC. With MSVC it only showed a warning (which is probably why this error went undetected at first):

```
C4002: too many arguments for function-like macro invocation
```

However, with GCC it straight up crashed, making it impossible to build.


cpinter:
Looks great, thank you very much!

Issue #293, opened 2025-12-04
Create a TOPAS dose engine


cpinter:
Create a TOPAS-dose-engine using as base
https://github.com/SlicerRt/SlicerRTExampleDoseEngine/ (We can offer
support on the TOPAS part eg via
https://github.com/topasmc/dicom-interface/)

cpinter:
Initial implementation has been integrated, closing the issue. For improvements and bugfixes new issues will be opened.

PR #328, opened 2026-06-22
ENH: Add TOPAS/Geant4 Monte Carlo dose engine

xskere:
Adds a new scripted dose engine that drives TOPAS simulations from within SlicerRT's External Beam Planning module.

Two beam source modes are supported:
- Simple beam: diverging proton source with configurable energy, field size, gantry/couch angles, and TsJaws collimation geometry.
- RT Ion Plan (TsRTIonSource): reads a DICOM RT Ion Plan and simulates all beams using the TOPAS dicom-interface. Requires the dicom-interface extension for TOPAS (https://github.com/topasmc/dicom-interface).

Beam parameters exposed in the UI include:
- Paths to the TOPAS installation directory, binary, and Geant4 data directory.
- Optional path to a DICOM RT Ion Plan file to enable TsRTIonSource mode.
- Particle type, beam energy, number of histories, ParticlesPerHistory downsampling factor, and CPU thread count.

Key implementation details:
- CT is exported as DICOM (or reused from Slicer's database to preserve Frame of Reference UID) and fed to TsDicomPatient.
- Beam placement uses the IsoCenter - ImgCenter convention matching TsRTIonComponents.
- Dose output is DICOM RT Dose, loaded via DicomRtImportExport.
- Multi-beam plans sum individual dose files with correct DoseGridScaling.
- Runs TOPAS natively on Linux or via WSL on Windows.

Also fixes dose window/level display in `qSlicerAbstractDoseEngine` and `qSlicerDoseEngineLogic` to fall back to the actual dose maximum when it differs from the prescription dose by more than 20%, preventing an invisible or clipped display when the simulated dose is not close to Rx.

Re #293


cpinter:
Looks great, thank you!

The only thing is I'd probably create a resource file for the Schneider data and simply copy it when needed, instead of writing it out from a long static string stored in-line.

@ferdymercury Please let us know what you think!

xskere:
> The only thing is I'd probably create a resource file for the Schneider data and simply copy it when needed, instead of writing it out from a long static string stored in-line.

Done! I changed it and now there is a HUtoMaterialSchneider.txt file in Resources and TOPAS gets it from there.

ferdymercury:
Thanks a lot!

> @ferdymercury Please let us know what you think!

Before merging, could you run it locally with this DICOM patient 
[Protons_01.zip](https://github.com/user-attachments/files/29252042/Protons_01.zip)
? To test that it's working fine.

The resulting DVH-comparison for the whole plan should look more or less like this:

image


If it doesn't look like it, then we need to crosscheck what parameter is missing wrt https://github.com/SebastiaanBreedveld/TROTS/blob/master/TOPAS_simulation/generate_scripts.py

xskere:
> Before merging, could you run it locally with this DICOM patient [Protons_01.zip](https://github.com/user-attachments/files/29252042/Protons_01.zip) ? To test that it's working fine.
> If it doesn't look like it, then we need to crosscheck what parameter is missing wrt https://github.com/SebastiaanBreedveld/TROTS/blob/master/TOPAS_simulation/generate_scripts.py

I used the TOPAS input files generated by scripts in the TROTS repository as a reference for creating the TOPAS input file generated by SlicerRT, so they should match. I'm going to run it and compare, once I have the results I will post here and we can go from there in case anything is wrong.

The reference against which you compared in that graph is the rtdose in the zip file? (To make sure I compare in the same way)

ferdymercury:
> The reference against which you compared in that graph is the rtdose in the zip file? (To make sure I compare in the same way)

Exactly. So the steps would be:

- Unzip patient1 (that already has a proper calibration of MU to number of protons, which is needed for TOPAS, and that 'hides' the range shifter, for simplicity).
- Load all files into DICOM - 3D Slicer database
- Set up TOPAS path via the GUI
- Set up a reasonable value of PPH so that it runs fast (1e5 or so)
- Click to run the three beams consecutively (or the whole plan if there is a button for that)
- Import the resulting .bin files into TOPAS and divide by the used PPH
- Sum the 3 generated TOPAS doses
- Click on DVH comparison plugin of SlicerRT, to compare imported DICOM rtdose.dcm file for the whole plan, with the sum of the three .bin files generated by TOPAS.

Thanks!!

ferdymercury:
> HUtoMaterialSchneider.txt file in Resources and TOPAS gets it from there.

It would be great if user can optionally select via GUI an alternative file from his own computer, since the schneider material is kinda  outdated (but still a good default).

xskere:
This is the result I got from SlicerRT:

dvh_comparison

xskere:
> It would be great if user can optionally select via GUI an alternative file from his own computer, since the schneider material is kinda outdated (but still a good default).

Ok, I will also add this to the Beam parameters

ferdymercury (approved):
Thanks!!

ferdymercury:
I had two follow-up comments:
- I think we are missing one button, where the user (optionally) selects a machine description file via file browser: see what I am referring to: https://github.com/topasmc/dicom-interface/wiki/using-machine-description-file
- question: can the user edit the TOPAS input files generated by Slicer, before running the simulation? Or everything happens on the fly?

xskere:
> * I think we are missing one button, where the user (optionally) selects a machine description file via file browser: see what I am referring to: https://github.com/topasmc/dicom-interface/wiki/using-machine-description-file

I will add it in the beam parameters, directly below the rtplan file, so it's visible.

> * question: can the user edit the TOPAS input files generated by Slicer, before running the simulation? Or everything happens on the fly?

No, the user can't edit it, the input file is generated at the moment of clicking 'Calculate dose', so other than the parameters the user may change in SlicerRT this can't be done.

So this is what can be changed (not directly on the input.txt but on SlicerRT)

- Without using the RTPlan file the user can change almost everything in the beam parameters.
- While using the RTPlan most parameters are overriden by the plan file (other than PPH), so only one beam with the RTPlan file specified in the beam parameters is needed (as well as the machine description file - if added)

ferdymercury:
Ok, thanks for the clarification!

xskere:
I added the machine description file to the beam parameters (same way as plan file is done)

ferdymercury (approved):
Thanks a lot!

cpinter:
Do you both consider this ready to merge?

xskere:
I'd say yes, I tested the last changes I made yesterday (adding the machine description file) and everything seems to be working correctly.

ferdymercury:
Yep, thanks!

PR #329, opened 2026-07-02
COMP: Fix IPOPT packaging in Windows

xskere:
`CPACK_INSTALL_CMAKE_PROJECTS` requires every listed directory to be a CMake build tree with a generated `cmake_install.cmake`. The Mumps/HSL/Ipopt entries pointed at autotools install prefixes (Unix) or a prebuilt binaries zip extraction (Windows), none of which are CMake projects, so package failed with a generic "Error when generating package: SlicerRT" whenever EXTENSION_BUILDS_IPOPT=ON.

Removed those three entries from `CPACK_INSTALL_CMAKE_PROJECTS` in the top-level CMakeLists.txt.
Added a proper `install()` rule in `ExternalBeamPlanning/Widgets/CMakeLists.txt` so the IPOPT DLLs are installed as part of that module (which is a real CMake target with a valid install script) instead.

With this fix, both build and package work on Windows and Linux.

Re #298

### DIFF
--- a/ExternalBeamPlanning/Widgets/CMakeLists.txt
+++ b/ExternalBeamPlanning/Widgets/CMakeLists.txt
@@ -281,6 +281,14 @@ if(EXTENSION_BUILDS_IPOPT AND WIN32 AND DEFINED Ipopt_DLL_DIR)
         "$<TARGET_FILE_DIR:${KIT}>"
     )
   endforeach()
+
+  # Also install the DLLs for packaging. Ipopt is a prebuilt zip on Windows (not a CMake
+  # build tree), so it cannot be registered via CPACK_INSTALL_CMAKE_PROJECTS; install it
+  # as part of this module instead.
+  install(DIRECTORY "${Ipopt_DLL_DIR}/"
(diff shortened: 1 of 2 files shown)


ExternalBeamPlanning/Widgets/CMakeLists.txt
```diff
@@ -281,6 +281,14 @@ if(EXTENSION_BUILDS_IPOPT AND WIN32 AND DEFINED Ipopt_DLL_DIR)
         "$<TARGET_FILE_DIR:${KIT}>"
     )
   endforeach()
+
+  # Also install the DLLs for packaging. Ipopt is a prebuilt zip on Windows (not a CMake
```
cpinter, line 285: Shouldn't we put this in an `if (WIN32)`? Not sure if this would break the packaging on Linux

xskere: It's inside `if(EXTENSION_BUILDS_IPOPT AND WIN32 AND DEFINED Ipopt_DLL_DIR)` so it should be already covered on that. Also I tested build + packaging on both Windows and Linux and it worked fine for both.

PR #335, opened 2026-07-04
STYLE: fix naming of HU to material file and rm duplicate setting

ferdymercury:
Schneider is just one option, not the name of the conversion process


cpinter:
Thanks! Please squash the commits into one.

Also please take note that the commit prefixes we use are these:
https://slicer.readthedocs.io/en/latest/developer_guide/contributing.html#how-to-write-commit-messages

Issue #333, opened 2026-07-03
Proton beams are kicked out of RTplan depending on previous state of IonPlan checkbox


ferdymercury:
When importing the dataset mentioned in https://github.com/SlicerRt/SlicerRT/issues/330

If you look at the Data Hierarchy, the three beams are correctly created inside the RTplan study.

<img width="539" height="222" alt="Image" src="https://github.com/user-attachments/assets/d79a9ca3-5adf-4472-b1d1-ab8d4a5c4280" />

Now, if I click on the "External Beam Planning" module, and the ion plan checkbox was not checked beforehand, the three beams are silently moved outside the RTplan study. It seems 3D slicer detects it's an ion plan and automatically checks the ion plan checkbox, but only after the beams have already been kicked out of the plan.

<img width="512" height="191" alt="Image" src="https://github.com/user-attachments/assets/5dd91be7-e249-432f-9008-7bbbe9ce472d" />

So I need to manually move them back inside the right hierarchy so that they appear in the ExternalBeam Planning module.

Or to remember having the ion plan checkbox active before selecting the RTPlan.

ferdymercury:
thanks for the PR!
Looks better now, however, if I by mistake unclick "Ion Plan", the beams are kicked out of the RTplan and 3 random sequences are created in the midst.

<img width="1335" height="360" alt="Image" src="https://github.com/user-attachments/assets/406f1182-aca3-4b50-bb89-e25a8ea2d0d5" />

I guess it would make more sense that when you inclick, it "unselects" instead the "Active RT plan" to nothing or to create a new folder display, rather than removing beams from a consistent RTplan study.
Or alternatively, that it pops up a warning dialog asking whether you want to continue removing these beams from the Plan.

Issue #331, opened 2026-07-03
Activating IonPlan checkbox disables TOPAS dose engine


ferdymercury:
When importing the dataset mentioned in https://github.com/SlicerRt/SlicerRT/issues/330
and clicking on Module ExternalBeam Planning, selecting the active RTplan automatically switches on the IonPlan checkbox, which automatically switches off TOPAS dose engine. It's not clear whether protons are also ions or not.

- TOPAS dose engine should only work for proton plans at the moment, rather than the other way round as currently (it only works if I unmark the checkbox).
- It could be convenient to have a tooltip popup when you hover the mouse over the checkbox
- Maybe it's better to have a dropdown menu to switch between X-rays, protons and other ions rather than an ion-plan checkbox ?

ferdymercury:
Looks better, thanks, but shouldn't then "Plastimatch proton" be also available in the dropdown menu when "Ion plan" is checked?

PR #338, opened 2026-07-08
BUG: Fix ion plan checkbox disabling TOPAS and losing beams

xskere:
- Let the TOPAS dose engine be selected for ion plans, since it already supports proton beams, and add a tooltip to the checkbox explaining that proton plans count as ion plans in DICOM
- Stop the checkbox from deleting all beams when it is merely being synced to reflect a plan that was already an ion plan on import, rather than toggled by the user
- Register dose engines and plan optimizers before refreshing their combo boxes when entering the module, so they are available on the first visit instead of only after re-entering the module

Re #331, #333


cpinter:
Looks good to me, thank you!

Issue #334, opened 2026-07-03
Misleading contour spacing calculation in vtkPlanarContourToRibbonModelConversionRule


ferdymercury:
With the dataset attached in https://github.com/SlicerRt/SlicerRT/issues/330, when importing CT+RTstruct, and then clicking on Segmentations, Representations, create Ribbon model, one gets the following unexpected warning:

```
Warning: In vtkPlanarContourToRibbonModelConversionRule.cxx, line 411
vtkPlanarContourToRibbonModelConversionRule (0x30ff9970): ComputeContourPlaneSpacing
Planar spacing: 4.9984. Frequency: 2.
```

I checked the original data and there are no missing contours, al contours are away 2.4992, so not sure why it calculates one slice with double the spacing?

The following is a minor comment/suggestion --> Besides, one gets these additional warnings:

```
Warning: In vtkPlanarContourToRibbonModelConversionRule.cxx, line 411
vtkPlanarContourToRibbonModelConversionRule (0x30ff9970): ComputeContourPlaneSpacing: Inconsistent plane spacing. Details:
  Planar spacing: 2.4991. Frequency: 7.
  Planar spacing: 2.4992. Frequency: 11.
```

From my point of view, these should be "Info or Debug" severity level rather than "Warning", because one cannot store RTstruct Z coordinates as decimal strings with more precision than that, so it's not possible to avoid those rounding errors. So I'd say if difference is less than 1 per mil, print info instead of warning.

lassoan:
Decimal string can be up to 16 bytes long (https://dicom.nema.org/medical/dicom/current/output/chtml/part05/sect_6.2.html), so a difference of 0.0001mm is not due to the limitation of the DICOM value representation, but a computation error in the writer software. But I agree that it is better not to flag it as error/warning if the difference is smaller than 0.1% of the distance between the two farthest contours.

ferdymercury:
> computation error in the writer software

Hmm I am using format_number_as_ds from pydicom so not sure why they don't use the 16bytes (https://github.com/SebastiaanBreedveld/TROTS/blob/f1d550570f15dbfd830cbd1576ecc9c256218b4c/DICOM_export/writeTROTSToDICOM.py#L9)

ferdymercury:
I just verified that the generated RTstruct dicom has the right interplance distance, with this script. I checked that all slices of that dataset are 2.4991836734693877 apart within 1e-9 precision.

[...]

PR #341, opened 2026-07-13
BUG: Fix incorrect ribbon model contour spacing warnings

xskere:
Creating a ribbon model from a segmentation could report inconsistent-spacing warnings, including gaps that looked twice as large as they should be. A fallback used while computing a contour's plane, which kicks in when the contour's initially chosen points happen to line up exactly straight, could end up reading coordinates from an unrelated contour instead of its own, corrupting the spacing calculation.

Spacing differences too small to be anything but precision limits of DICOM decimal string coordinates are also now logged at a lower severity instead of as a warning, since they are not actionable.

Re #334


cpinter:
Thank you, looks good!

PR #343, opened 2026-07-22
BUG: Fix four failing tests

xskere:
Four tests were failing on SlicerRT. In every case the test itself had gone stale against changes elsewhere, none were production bugs. Each fix is a separate commit that names the change responsible.

1. DicomRtImportTest:
Asserted a fixed subject-hierarchy item count last set in 2022. The RTIon plan/beam loading improvements from August 2025 (commit 37089a96) added a dose reference table under the plan, raising the real count by one. Bumped the expected value.

2. Isodose logic test (EclipseProstate):
The isosurface volume drifted ~0.4% past tolerance. The decimation step was intentionally removed from isosurface generation in commit a2b2967f (November 2024, "Add real-time mode to Isodose computation") because it broke very small surfaces, but the baseline surface was never regenerated. Recomputed the baseline against the current pipeline.

3. Rooms Eye View logic test:
Expected 13 transform nodes, found 15. The external IEC transform library this extension builds against added two coordinate systems for Monte Carlo and proton transport support in February 2025; the transform hierarchy creates a node for every pair the library reports, so the real count grew to 15. Bumped the expected value, and added scene teardown at the end of the test to silence a pre-existing leak report on the clean-exit path.

4. Plastimatch proton dose engine test:
- It called the old single-target setter, replaced by a multi-target one in commit 1838fb87 (June 2026), so it aborted immediately.
- It never defined the engine's beam parameters. Commit 6b66601a (October 2024) moved that step to the point where a beam parameters widget is registered, which only happens when the External Beam Planning module is opened. A headless run opens no module, so dose was computed with missing parameters and failed. The test now registers a widget itself, matching what the application does on startup.
- The expected dose values predated both the Plastimatch upgrade to 1.10 (commit efd3573b, November 2024) and the beam source placement correction (commit 4a33bea0, October 2024).


cpinter:
Excellent, thank you very much!

Issue #330, opened 2026-07-02
Importing DICOM CT + RTstruct + RTplan simultaneously leaves 3D view in a weird state


ferdymercury:
I have a DICOM dataset [Protons_01.zip](https://github.com/user-attachments/files/29604922/Protons_01.zip) with :
- CT
- RTstruct
- RTplan

On Slicer-5.13.0-2026-06-30 (but also in stable version) with SlicerRT plugin installed, if I go to the DICOM load module, and then import simultaneously just CT + Rtstruct, and then on the 3D pane I click on "Center view", I see the patient 3D surface.

Instead, if i import simultaneously the three files, and then on the 3D pane I click on "Center view", the L-P-A-S bounding box suddenly disappears and the 3D screen is just blue.

Besides, I get the following warnings that should not appear, maybe they are helfpul for finding the bug:

```
vtkSlicerSubjectHierarchyModuleLogic::InsertDicomSeriesInHierarchy: Subject hierarchy item with DICOM UID '1.2.826.0.1.3680043.8.498.84222194691086137828904549844681123832' cannot be found
vtkSlicerDicomRtImportExportModuleLogic::InsertSeriesInSubjectHierarchy: Failed to insert series with Instance UID 1.2.826.0.1.3680043.8.498.84222194691086137828904549844681123832
SetupRtImageGeometry: Failed to get RT Plan DICOM UID for beam '1'
SetupRtImageGeometry: Failed to get RT Plan DICOM UID for beam '2'
SetupRtImageGeometry: Failed to get RT Plan DICOM UID for beam '3'
GetParentPlanNode: Failed to access subject hierarchy node
CreateRangeShifterPolyData: Invalid parent ion beam node
CreateRangeShifterPolyData: Invalid parent ion beam node
```

If I do the process in two steps, first import CT+RTstruct, click on Center, then import RTplan, it's fine, but if I click on "Center" again, it disappears.

If I delete the RTplan by hand in the left Data pane, the Center view button starts to work again.

In summary: having an RTplan loaded should not make the 3D view disappear.

This is happening for me in Ubuntu22.04, maybe in Windows it's different.

Thanks in advance!

cpinter:
Can you please post a screenshot?

ferdymercury:
Thanks for the reply!

Sure, this is simultaneous import:
[...]

PR #344, opened 2026-07-24
BUG: Fix range shifter placement and 3D centering on ion plan import

xskere:
Two problems left imported proton plans with their range shifters in the wrong place and the 3D view unable to center on the patient:

First, the dynamic beam sequence import applied the isocenter translation twice: once while updating the beam transform and again explicitly afterward. This placed every imported beam and its range shifter at twice the isocenter offset, more than a metre from the patient. The redundant second translation has been removed so imported beams and range shifters sit at their correct location.

Second, a retracted range shifter was drawn shifted sideways off the beam axis, which pushed it outside the field of view and further skewed the automatic centering. The range shifter model now always stays on the beam axis, and whether it is retracted or inserted is conveyed through its transparency rather than it being moved out of position.

Re #330

### DIFF
--- a/Beams/MRML/vtkMRMLRTIonRangeShifterNode.cxx
+++ b/Beams/MRML/vtkMRMLRTIonRangeShifterNode.cxx
@@ -257,15 +257,18 @@ void vtkMRMLRTIonRangeShifterNode::CreateRangeShifterPolyData(vtkPolyData* range
   double isoToRsDistance = ionBeamNode->GetIsocenterToRangeShifterDistance();
   double wet = ionBeamNode->GetRangeShifterWET();
   const char* setting = ionBeamNode->GetRangeShifterSetting();
-  double beamAxisOffset = 0.; // in/out beam offset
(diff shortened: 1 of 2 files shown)


cpinter:
If @ferdymercury agrees then we can integrate. Thank you!

ferdymercury (approved):
Thanks a lot!

Not sure if the RangeShifter Opacity settings could be configured via two GUI buttons ? Or is hard-coded 0.05 vs 0.3 good enough ?

xskere:
> Not sure if the RangeShifter Opacity settings could be configured via two GUI buttons ? Or is hard-coded 0.05 vs 0.3 good enough ?

From what I tested, using 0.05 vs. 0.3 opacity made it easy to tell when it was ON or OFF. That said, we could add a UI control for adjusting opacity later if needed, but we'd first need to determine where it should go in the interface.

ferdymercury:
> From what I tested, using 0.05 vs. 0.3 opacity made it easy to tell when it was ON or OFF.

sounds good to me then to leave it as is for the moment.

PR #345, opened 2026-07-24
Fix ion plan checkbox destroying beams and enable proton engine for ion plans

xskere:
First, this adds a confirm before the ion plan checkbox deletes beams:

Switching a plan between ion and conventional deletes every beam it contains, because the two kinds of beam are different types of object that cannot be converted into one another. Previously this happened silently and could not be undone, on a checkbox that is easy to hit by accident, and unchecking it on an imported plan scattered stray sequence items across the scene.

- The user is now asked to confirm whenever the change would discard beams. Declining leaves both the plan and the checkbox as they were.
- Removing a beam that was loaded from a DICOM dynamic beam sequence now also removes the sequence browser and sequences that drive it. Before, only the beam was removed, so the sequences module recreated it at the top of the scene, outside of any plan, once per dynamic beam. This covers every way of deleting a beam, not only the checkbox.

Second, keeps the Plastimatch proton engine available for ion plans:

Turning on the ion plan option removed the proton dose engine from the list of choices. It is now offered for ion plans as well.

The hook that lets an engine adjust its beam parameters when the ion plan option changes previously logged an error for any engine that did not implement it; it is now a silent no-op, since only engines with plan-dependent parameters need to react.

Re #331, #333

### DIFF
--- a/Beams/MRML/CMakeLists.txt
+++ b/Beams/MRML/CMakeLists.txt
@@ -9,6 +9,7 @@ set(${KIT}_INCLUDE_DIRECTORIES
   ${Slicer_Base_INCLUDE_DIRS}
   ${MRMLCore_INCLUDE_DIRS}
   ${vtkSlicerSegmentationsModuleLogic_INCLUDE_DIRS}
+  ${vtkSlicerSequencesModuleMRML_INCLUDE_DIRS}
   )
 
 # --------------------------------------------------------------------------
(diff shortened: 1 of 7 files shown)


cpinter:
Looks good, thank you!

@ferdymercury Please let us know what you think.

ferdymercury (approved):

ferdymercury:
Looks great thanks!

PR #346, opened 2026-08-21
BUG: Fix multi leaf collimator workflow in beam parameters

xskere:
The table selector was only given a scene when the beam already had a table and was left disabled otherwise, so a beam without one had a selector that could never be used. Generating a table then created it in the scene but could not select it, leaving it unattached to the beam, and the buttons acting on it stayed greyed out. The selector is now filled for every beam, a generated table is attached to the beam directly rather than through the selection, and the buttons follow the table the beam actually carries. Updating the boundaries of an existing table announces the change as one edit, so the beam model is rebuilt from the finished values instead of keeping its old shape. The scan spot selector had the same defect and is fixed the same way.

When the leaf positions were calculated, only the pairs covering the calculated outline were given positions, so pairs opened for a target selected earlier stayed open and the opening covered the old target and the new one together. Every pair is now closed before the ones the new outline needs are opened, and the pair count is derived so a table without rows yields no pairs rather than a huge number.

The closed curve holding the calculated opening is created hidden, being a by-product of the calculation rather than something to look at, and it is no longer released by hand after being taken out of the scene, which freed it twice when the target collapsed to a line or a point.

### DIFF
--- a/Beams/Logic/vtkSlicerMLCPositionLogic.cxx
+++ b/Beams/Logic/vtkSlicerMLCPositionLogic.cxx
@@ -91,6 +91,8 @@ namespace
 
 const char* MLCX_BOUNDARYANDPOSITION = "MLCX_BoundaryAndPosition";
 const char* MLCY_BOUNDARYANDPOSITION = "MLCY_BoundaryAndPosition";
+// Position given to both leaves of a pair that is shut, as used for a newly generated table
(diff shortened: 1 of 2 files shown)


cpinter:
Looks good to me, thank you! Just pinging @ferdymercury in case he wants to look at it (may find something odd). Thank you!

